PR #15, opened 2020-03-10
Fixes #14 - restore stylelint dollar-variable-default

### LINKED ISSUE #14: Restore stylelint config to disable "dollar-variable-default" rule
**Describe the bug**
When compiling arizona-bootstrap locally (npm), all SCSS variables (116 total, currently) error out and ask for the `!default` flag to be applied. This prevents local work from compiling. This was accidentally removed from **.stylelintrc** in #13; this PR will add it back in.

**To Reproduce**
1. `cd` into arizona-bootstrap local repo
2. `npm run start` (or other method)
3. Make a change to the `_custom-variables.scss` file
4. Save
5. Wait for re-compile, and list see errors pertaining to $ variables

**Expected behavior**
Compile without `Expected !default flag for $variable-name` errors

akslay:
Restores the stylelint config to disable "dollar-variable-default"

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -11,7 +11,7 @@
       "border-bottom-left-radius",
       "transition"
     ],
+    "function-blacklist": ["calc"],
     "scss/dollar-variable-default": null
-    "function-blacklist": ["calc"]
   }
 }
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -11,6 +11,7 @@
       "border-bottom-left-radius",
       "transition"
     ],
+    "scss/dollar-variable-default": null
     "function-blacklist": ["calc"]
   }
 }


.stylelintrc
```diff
@@ -11,6 +11,7 @@
       "border-bottom-left-radius",
       "transition"
     ],
+    "scss/dollar-variable-default": null
```
joeparsons: One _very minor_ nitpick: I noticed that in the "upstream" TWBS stylelit config has the scss/dollar-variable-default rule last (after the function-blacklist rule):
https://github.com/twbs/bootstrap/blob/master/.stylelintrc

So maybe it would make sense to keep ours in that order, too?  (Just to make it easy to track future upstream changes)

akslay:
> Sorry about that change: the main source of false positives in the linting failures.

No problem at all Martin. (:

joeparsons:
Would anyone be opposed to merging this right away?  I think it blocks the other work @akslay is doing.

PR #16, opened 2020-03-11
Fixes #9 - Finalize Bootstrap audit

### LINKED ISSUE #9: Finalize AZ Bootstrap
https://jira.arizona.edu/browse/UADIGITAL-2187

Finish going through audit items:https://docs.google.com/spreadsheets/d/1tgGDGHV6Wqts3b88Jp_CbmsUnmMg1O5s0W7zUgeXuLs/edit#gid=0

- Going back through the audit results to finish out "to-do" items and auditing the _custom_overrides.scss file.

akslay:
Creating a pull request so that we have something to look at for Friday's meeting. Still need to go through the audit spreadsheet again to make sure we have everything we want for the initial release of AZ Bootstrap


akslay:
Audit spreadsheet: https://docs.google.com/spreadsheets/d/1tgGDGHV6Wqts3b88Jp_CbmsUnmMg1O5s0W7zUgeXuLs/edit#gid=0

akslay:
- [ ]  Decision on adding back in Badge Pills

akslay:
> Most text color options on the .bg-transparent colors do not pass color contrast checking (using contrast-ratio.com). There are also issues with other combinations, such as bg-success with text-white. I'd suggest not enabling combinations that don't meet adopted standards.

I'm not sure there's really a way to "disable" color combinations. And I agree transparency is definitely an issue with regards to color contrast. But I think that might be a bigger discussion item overall since it's something we're including specifically for use with AZQS components that have text over background images ("Marquee"/"Media Row").

joshuasosa:
> I'm not sure there's really a way to "disable" color combinations.

If you can't do it as-is with the SCSS, perhaps be more explicit with their inclusion in the SCSS. Instead of looping through all possible bgs and adding all possible text-color options, don't do that and explicitly define each text-color option that works fine with each possible bg.

> And I agree transparency is definitely an issue with regards to color contrast. But I think that might be a bigger discussion item overall since it's something we're including specifically for use with AZQS components that have text over background images ("Marquee"/"Media Row").

The site I used, contrast-ratio.com, says:
>When semi-transparent colors are involved as backgrounds, the contrast ratio will have an error margin, to account for the different colors they may be over.

I read that as meaning it does try to account for text over different background colors. With the background images, it may be even more important to ensure combinations meet standards since certain bg graphics can make text harder to read.

akslay:
I'm just looping through it so that whenever there's a `.text-[color]` class inside a `div` with a background color, those text color classes aren't being overwritten by the "default" color, which is accessible. I added in a "default" (black or white) text color just so the initial color is accessible when using a background color. After which point it's up to the content editor or developer to explicitly choose which color they're wanting to use inside that background color. 

Rather than allowing or disallowing color combinations, I think I prefer the route of showing them which options are accessible (the table/chart we have) and recommending which color options to use via that chart. But that's just my personal viewpoint on the matter, so that might be another good discussion item for Friday's meeting. (: 

And if we do go the route of defining which text colors are allowed inside which background colors, I think that might be a good opportunity for a follow-up improvement after V1, since V1 is really just an MVP/trying to get it out the door, and not a finished product.

joshuasosa:
I strongly prefer the route of not even including combinations that aren't accessible. It doesn't make any sense to include them, except to purposefully be non-compliant. By including a selector like .bg-transparent-red p.text-red, you are implicitly saying it's OK to use it because it's there and the selector is specifically defined that way. But that combination doesn't look like it would pass in any scenario, regardless of the color behind it:
https://contrast-ratio.com/#%23ab0520-on-rgba%28171%2C%205%2C%2032%2C%200.7%29

I would also rather not include something we **know** to be non-compliant than back-track on it at a later date.

scss/_mixins.scss
```diff
@@ -25,17 +16,84 @@
...
 
-//
-// > TRANSPARENT BACKGROUND COLORS
-//
+/*
+* > BACKGROUND COLORS
+*/
+
+// >> Transparent Backgrounds
 @mixin background-transparent($bg-color) {
```
trackleft: This mixin can be included in the separate PR. Sorry for any inconvenience

scss/_custom-styles.scss
```diff
@@ -31,319 +31,50 @@
...
-      }
-    }
   }
 }
 
 // >>> Brand Transparent
 @each $color, $value in $colors-transparent {
   .bg-#{$color} {
+    @include text-color-contrast($value);
     @include background-transparent($value);
```
trackleft: Should we remove this line for now, and include it in a separate PR, where we can spend more time figuring out which translucent colors we want?

joshuasosa:
> I think we can safely remove background transparency classes in this PR, and include them in another, much like the pills PR.

That'd probably make things easier to get this PR in, since most issues I was seeing are with the transparent classes. There are, however, other combinations that should be addressed. For example, .bg-success is not a transparent class, and .bg-success p.text-white is not an accessible combination.
https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=70B865

All combinations should likewise be checked.

joeparsons:
@akslay Will add some minor syntax changes to resolve remaining linting errors.

joeparsons:
@trackleft will address remaining color combination issues in #22.

PR #23, opened 2020-03-20
Bug/22 remove unneeded color combination classes

### LINKED ISSUE #22: Remove unnecessary color class combinations
**Describe the bug**
Existing scss creates unnecessary/misleading color class combinations.

e.g. Remove unnecessary combinations like this: 
```.bg-red p.text-color-red```

### PR DESCRIPTION
**Pull request by trackleft**
Remove unneeded color combination classes.

## Description
Fixes #22 by changing the order in which text-colors are added to the css files, bg classes first then text colors.

## Related Issue
#22 

## How Has This Been Tested?
Locally

## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

### CODE REVIEW

File: scss/_mixins.scss [hunk 1]
```diff
@@ -146,17 +91,4 @@
 */
 @mixin callout-variant($color) {
   border-left-color: $color;
-
```
trackleft: Not sure about this part, but took it out anyway.

akslay: This part makes sure that the Callouts have the matching text color as the border. And anything using an inaccessible/light color (like Sky) uses the Dark Silver heading color. So with this change, all the headings are Dark Silver instead of the correct color.

trackleft: Thanks I've added back

scss/_custom-styles.scss
```diff
@@ -83,6 +55,34 @@
 }
```
trackleft: Moved text color beneath background color.

scss/_mixins.scss
```diff
@@ -29,64 +29,10 @@
 @mixin text-color-contrast($bg-color) {
   @if $bg-color == $bloom or $bg-color == $white or $bg-color == $sky or $bg-color == $oasis or $bg-color == $cool-gray or $bg-color == $warm-gray or $bg-color == $silver or $bg-color == $leaf or $bg-color == $warning or $bg-color == $light or $bg-color == $gray-100 or $bg-color == $gray-200 or $bg-color == $gray-300 or $bg-color == $gray-400 or $bg-color == $gray-500 {
     color: $black;
-    p,
```
trackleft: Changing the order allowed me to remove all of the specific selectors.

scss/_typography.scss
```diff
@@ -48,6 +48,7 @@ h6,
   margin: $heading-margins;
   font-family: $font-family-sans-serif;
   color: $dark-silver;
+  color: inherit;
```
akslay: Nice 👍🏻This will make sure headers are accessible by default when there's a background color.

scss/_mixins.scss
```diff
@@ -28,65 +28,13 @@
...
-      }
-      @each $text-color, $text-value in $theme-colors {
-        // ensure elements using color classes aren't overwritten
-        &.text-#{$text-color} {
-          color: $text-value;
-        }
-      }
-    }
+    // stylelint-disable-next-line declaration-no-important
+    color: $black !important;
```
joshuasosa: Do these need to be important?

scss/_typography.scss
joshuasosa: Are both color properties necessary? The lower one will just override the color: $dark-silver one.

trackleft:
@joshuasosa https://github.com/az-digital/arizona-bootstrap/pull/23#discussion_r395949001
Here was my thought process that turned out to not work at all, even though I could have sworn I saw it working at one point.

I believe the first one is a fallback for the second one

Currently the default text color for headings is set to dark-silver.
```
h3,
h4,
h5,
h6, .h3, .text-size-h3, .h4, .text-size-h4, .h5, .text-size-h5, .h6, .text-size-h6 {
  margin: 1.042em 0 0.667em;
  font-family: MiloWeb, Verdana, Geneva, sans-serif;
  color: #49595e;
  color: inherit;
}
```

Further down the css file, we then set headings to an accessible color depending on their background with a less specific declaration for example with panels. To be honest, I am not 100% sure these ones have to be set to `!important` since they should influence anything within them.

```
.bg-gray-400 {
  color: #000 !important;
  background-color: #ced4da;
}
```
Dark silver does not always pass WCAG AAA, but Black does.


Then, let's say I want to set `bg-gray-400` headings to midnight instead of black since it also passes, i would add the 
`text-midnight` class which is further down the css file. I believe these do need to be set as `!important` and BS4 also sets color only classes to `!important`
```
.text-midnight {
  color: #001c48 !important;
}
```

Turns out inherit just overrides the set color, and since we have to body set to a specific color the body text color is inherited.  Back to the drawing board.

trackleft:
For the accessible headings `@mixin` I think we need to scrap it for now, unless someone can figure out way to make it work without adding thousands of lines of additional CSS.

trackleft:
The issue that we are running into is the default heading color of dark-silver

Maybe we should remove the default color for headings and just rely on the body text color, so that we can override more easily where it needs to be overridden.


CSS for default heading colors is here: 
```
h3,
h4,
h5,
h6, .h3, .text-size-h3, .h4, .text-size-h4, .h5, .text-size-h5, .h6, .text-size-h6 {
  margin: 1.042em 0 0.667em;
  font-family: MiloWeb, Verdana, Geneva, sans-serif;
  color: #49595e;
  color: inherit;
}
```

Since background colors are usually set on parents of headings.

For example

``` 
.bg-red {
background-color: red;
color: white;
}
```
<div class="bg-red">
  <h3>Text gets color from h3 color setting instead of bg-red color setting.</h3>
</div>
```

File: scss/_mixins.scss [hunk 6]
```diff
@@ -73,27 +61,15 @@
     .h4,
     .h5,
     .h6 {
```
trackleft: May be able to replace lines 51 - 63 with a * , but not sure if that is too heavy handed.

scss/_mixins.scss
```diff
@@ -42,19 +42,7 @@
...
```
trackleft: May be able to replace lines 32 - 44 with a * , but not sure if that is too heavy handed.

joshuasosa: I don't think * is necessary. Even p on lines 32 and 51 don't seem necessary. It's only the headers that are explicilty changed in _typography.

Issue #24, opened 2020-03-27
The arizona-bootstrap dev dependencies need re-synchronized with the v4-dev branch of upstream bootstrap


mmunro-ltrr:
**Describe the bug**
Upstream https://github.com/twbs/bootstrap/tree/v4-dev now includes another set of dev dependency changes, in particular from the pull request https://github.com/twbs/bootstrap/pull/30006 ( I like the phrase in the initial comment “In the past we considered such changes breaking ones.“). For example, there is a move to a new major release version of Jekyll.

**To Reproduce**
View the arizona-bootstrap repository on GitHub: there is now a security warning which does not appear on the upstream repository.
Errors appear in some local development environments.

PR #51, opened 2020-04-13
Feature/10 add proxima nova font

trackleft:
As a first step, we can merge this, however the font is attached to my account on typekit.

## Tasks

- [x] determine if we need to be more specific with fallback font options or if native to browser/os fallback fonts get the job done
- [x] merge master into this branch to ensure we're testing with latest/greatest
- [ ] extra special read / sets of eyes on content on https://review.digital.arizona.edu/arizona-bootstrap/feature/10-add-proxima-nova-font/docs/0.0/content/font/


barrettbaffert:
I think it’s fine to move forward with it attached to Chris’ typekit profile so long as everything else looks as expected. We are fairly confident that the ultimate solution (Creative Cloud Libraries?) will simply replace the url that the font is sourced from. Right @trackleft?

trackleft:
@barrettbaffert yes, the ultimate goal will be to use a font asset with distributed governance, and as far as arizona-bootstrap usage is concerned it should be a seamless transition from this solution to the ultimate solution as long as they stay up to date with arizona-bootstrap in their markup.  At this time I don't think we should recommend anyone using my link without arizona-bootstrap.

akslay:
I don't recall if I asked this before (if I did, sorry): since we're switching to this font, should we remove the `.sans` class and also add it to the deprecated section with no backwards compatibility? So that we're not loading the old font for the sake of backwards compatibility.

mmunro-ltrr:
I was checking to see if there were any potential conflicts with the changes needed for Issue #24 (the dev dependency updates). So far ok.

site/docs/0.0/examples/album/index.html
```diff
@@ -1,7 +1,7 @@
 ---
 layout: examples
 title: Album example
-extra_css: "album.css"
+<!--extra_css: "album.css"-->
```
joeparsons, line 4: I don't think it's valid to use HTML comments like these within the [YAML front-matter](https://jekyllrb.com/docs/front-matter/).  I'm getting errors like this when I try running `npm run docs-compile` locally:

```
YAML Exception reading /arizona-bootstrap/site/docs/0.0/examples/blog/index.html: (<unknown>): could not find expected ':' while scanning a simple key at line 4 column 1
YAML Exception reading /arizona-bootstrap/site/docs/0.0/examples/album/index.html: (<unknown>): did not find expected key while parsing a block mapping at line 2 column 1
```

Maybe we should just delete these lines instead?

joeparsons: Or alternatively comment these lines with `#`.

site/docs/0.0/content/font.md
```diff
@@ -0,0 +1,746 @@
...
+
+  Before using the Milo font suite, please complete the <a href="https://brand.arizona.edu/font-license-agreement" target="_blank">font license
+  agreement</a> <span class="label
+  label-warning">Important</span>
+
+  For projects not using UA Bootstrap, follow these instructions on <a href="https://brand.arizona.edu/how-add-milo-your-website" target="_blank">how to add Milo to your website</a>.
+  Be aware that [UA Bootstrap reference links](index.html#ua-bootstrap-reference-links) come "pre-loaded"
+  with the Milo font suite. In other words, if you're using UA Bootstrap, font
+  installation instructions are not required.<span class="label
+  label-warning">Important</span>
```
joeparsons, line 106: Do we really still want to reference Milo / UA Bootstrap here?

site/docs/0.0/content/font.md
```diff
@@ -0,0 +1,746 @@
...
+team manages a typekit webfont project that is referenced in Arizona Bootstrap.
+
+#### Notes to consider when using the centrally mananged webfont project.
+Our license with typekit allows anyone with a netid to create their own webfont
+project in creative cloud typekit and use that in all of their web projects. Of
+course there are pluses and minus to each approach.
+
+#### Managing your own project
+
+**Pluses**
```
joeparsons, line 22: Nitpicky/style suggestion: Maybe "Advantages" / "Disadvantages" instead of "Pluses" / "Minuses"?

site/docs/0.0/content/font.md
trackleft: Looks like I left a lot of stuff in there on accident

site/docs/0.0/content/font.md
trackleft: Changed to "Advantages" / "Disadvantages"

site/docs/0.0/content/font.md
trackleft: Removed all of that.

site/docs/0.0/examples/album/index.html
trackleft: Commented with `#`

barrettbaffert:
Looking good! 

When I checkout this branch and `npm i && npm run dist && npm run start` I see the docs with Proxima Nova (as I'd expect!). I did grep for the word "milo" and nothing of note came up.

I do see this error when I `npm run start` this branch, but not when I run master

```npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! arizona-bootstrap@0.0.2 css-lint-docs: `stylelint "site/docs/**/assets/scss/*.scss" "site/docs/**/*.css" --cache --cache-location .cache/.stylelintcache````

File: site/docs/0.0/content/font.md [hunk 4]
```diff
@@ -0,0 +1,749 @@
...
+## How to Use
+  The Proxima Nova font suite is available for official use by University of Arizona
+  employees through a license with Adobe Typekit.
+
+### Reference link
+  Arizona Bootstrap uses a Creative Cloud Library with shared governance to
+  manage our font reference links within Arizona Bootstrap. Alternatively, place
+  this line of code at the top of your main stylesheet.
+
+  ```@import "https://use.typekit.net/tgs5qmz.css";```
```
barrettbaffert, line 100: @trackleft can you please add...

- [ ]  WARNING this shouldn't be used

barrettbaffert: we don't yet support wide usage of the font from a centralized reference link

trackleft:
FYI, It's actually considered worse to link the fonts in the stylesheet since it blocks the rest of the stylesheet from downloading.

In summary `@import` within the stylesheet = BAD
` <link href='' type='text/css'>` = GOOD

site/docs/0.0/content/font.md
trackleft: Added a BEWARE! alert.

trackleft:
> @barrettbaffert yes, the ultimate goal will be to use a font asset with distributed governance, and as far as arizona-bootstrap usage is concerned it should be a seamless transition from this solution to the ultimate solution as long as they stay up to date with arizona-bootstrap in their markup. At this time I don't think we should recommend anyone using my link without arizona-bootstrap or with a copy of arizona-bootstrap that they do not intend on keeping up to date as we release new versions.

This is no longer the case, people will have to update the proxima nova link

trackleft:
FYI https://github.com/stylelint/stylelint/issues/4154

mmunro-ltrr:
Obviously this now needs re-synchronized with the current Arizona Bootstrap version, but it's getting to be a high priority to merge, since the front end group playing around with pieces of Arizona Bootstrap in various contexts will be doing more in the next few weeks. I'm actually using Arizona Bootstrap on one production site, & would like to have the dramatic shift in typography happen sooner rather than later!

.gitignore
```diff
@@ -2,6 +2,7 @@
 /_site/
 /site/.jekyll-cache/
 /site/.jekyll-metadata
+/site/.jekyll-cache
```
joeparsons, line 5: Could this be combined with line 3?

Issue #58, opened 2020-04-16
MVP GitHub Actions for AZ Bootstrap Pull Requests


joeparsons:
## User Story(s)
As a _member_ of the Arizona Digital team, I'd like to have a review site automatically generated and deployed when I submit PR in order to make it easier for _members_ to review my proposed changes.
…

## Conditions of satisfaction
Should add GitHub Actions configuration that does the following for submitted Pull Requests:
- [x] Builds documentation site artifact
- [x] Deploys documentation site artifact to review site
- [ ] Adds link to review site on PR

PR #61, opened 2020-04-22
#58: WIP Docker container GitHub action.

joeparsons:


mmunro-ltrr:
I'm rethinking how we're using Docker, both in GitHub actions and in local dev environments. On the one hand it would be nice to speed up the `npm install` step by having a cached copy of node_modules. In GitHub's Node.js-based workflows there are nice features for caching various things,
https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows
but these aren't available inside the containers of a Docker-based workflow, and of course wouldn't be reusable in local dev environments. I think it's worth having a spike to sort this out, & on a fairly short timescale (days, not weeks). The other background issue (not an Issue yet!) is that if we're really following the v4-dev branch of twbs/bootstrap we shouldn't really be copying the js assets into arizona-bootstrap from an older tagged release, but rolling our own, & that would be a real performance hit if we were doing it on every build, so having our own cached copy of the build artefacts would be good for that, but with some mechanism for clearing the cache if something in a pull request or local edit changed the dependencies.

joeparsons:
May revisit this later.

PR #80, opened 2020-05-08
Fixes #79 - Remove h3 from accordion example

### LINKED ISSUE #79: Remove h3 from accordion example
## Problem/Motivation
Collapse example on the documentation site is using `h3`s, which is causing margin issues.

### Describe the bug
Margin inside collapse headers, which doesn't need to be there.
Screen Shot 2020-05-08 at 10 23 39 AM

### To Reproduce
View the accordion example on the docs site, note the margin-top on the header due to the use of an `h3` tag.

## Proposed resolution
Remove the `h3` from the documentation example.

### Expected behavior
No margin.

akslay:
Fixes #79 by removing the h3 from the accordion example, which removes the excess margin that shouldn't be there.

### DIFF
--- a/site/docs/0.0/components/collapse.md
+++ b/site/docs/0.0/components/collapse.md
@@ -74,38 +74,38 @@ Multiple `<button>` or `<a>` can show and hide an element if they each reference
 Using the [card]({{ site.baseurl }}/docs/{{ site.docs_version }}/components/card/) component, you can extend the default collapse behavior to create an accordion. To properly achieve the accordion style, be sure to use `.accordion` as a wrapper.
 
 {% capture example %}
-<div class="accordion" id="accordionExample">
+<div class="accordion" id="ex1_accordionExample">
   <div class="card">
-    <div class="card-header" id="headingOne">
-      <button class="btn btn-link" type="button" data-toggle="collapse" data-target="#collapseOne" aria-expanded="true" aria-controls="collapseOne">
+    <div class="card-header" id="ex1_headingOne">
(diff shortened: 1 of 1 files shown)


akslay:
Bug: https://review.digital.arizona.edu/arizona-bootstrap/master/docs/0.0/components/collapse/#accordion-example

Fix: https://review.digital.arizona.edu/arizona-bootstrap/80/merge/docs/0.0/components/collapse/#accordion-example

site/docs/0.0/components/collapse.md
```diff
@@ -91,11 +89,9 @@ Using the [card]({{ site.baseurl }}/docs/{{ site.docs_version }}/components/card
   </div>
   <div class="card">
     <div class="card-header" id="headingTwo">
-      <h3 class="mb-0">
```
trackleft, line 94: I think we should add a text-size-h3 class

trackleft: Also, we should consider how screen readers and other semantics will be affected..

trackleft: mx-0 maybe?

joeparsons:
Some background about a TWBS change that was made that changed the headings within Collapse examples to h2 elements:
https://github.com/twbs/bootstrap/issues/27619

akslay:
- [ ] Add a second example including the use of headings
- [ ] Include verbiage about "depending on your use case, you may or may not need to include headings in your accordion markup"

site/docs/0.0/components/collapse.md
akslay: Adding back in `h3`s for the second example (follow-up items)

trackleft:
RTBC IMHO https://review.digital.arizona.edu/arizona-bootstrap/bug/79/docs/0.0/components/collapse/#accordion-example

PR #83, opened 2020-05-29
Bootstrap Release Action

kevcooper:
This PR adds a github action that facilitates creating a new release of Arizona Bootstrap.

To trigger this action, update the version within package.json and package-lock.json by running `npm version 0.0.3` and opening a PR with the updates. Once the PR is merged to master, a new release will be generated based on the version number specified in package.json.

Open questions:

- [ ] Do we still want to go "back to dev" after a release and append `-dev` to the version numbers

- [ ] Which release types do we want to be "pre-release"? (alpha, beta, etc)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,6 +20,7 @@ on:
   push:
     branches:
       - master
+      - user-story/*
     paths:
       - package.json
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -55,6 +55,8 @@ jobs:
         run: |
           jq --arg suffix '-dev' '.version |= . + $suffix' package.json > package.json.new
           mv package.json.new package.json
+          jq --arg suffix '-dev' '.version |= . + $suffix' package-lock.json > package-lock.json.new
+          mv package-lock.json.new package-lock.json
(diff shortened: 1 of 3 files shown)


joeparsons:
Just looked at a few popular npm packages (react, express, etc) and it looks like they all follow the convention of only updating the `version` value in `package.json` when a new release is tagged and the version number stays the same for all commits between releases.

joeparsons:
Todos (could potentially be in follow-up PRs):

- [ ] Document instructions for creating a release PR (in README?)
- [ ] Add step to release workflow for triggering arizona-bootstrap-packagist repo

.github/workflows/release.yml
```diff
@@ -0,0 +1,63 @@
...
+        run: |
+          exit 1
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: "v${{ env.AZ_VERSION }}"
+          release_name: "v${{ env.AZ_VERSION }}"
```
joeparsons, line 50: I think we actually want to leave to `v` off of the `release_name`.

joeparsons: NPM's convention seems to be adding the `v` only to tag names.

PR #85, opened 2020-05-29
0.0.3

kevcooper:
This is a test of the new release process

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "arizona-bootstrap",
-  "version": "0.0.2",
+  "version": "0.0.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


joeparsons:
Shouldn't the `dist` directory also be updated?

kevcooper:
If i run `npm run dist` there are no changes to the dist directory. If I first delete `dist` before running `npm run dist`, I get a bunch of npm errors related to being unable to move non-existant directories.

Should there currently be changes that show up in dist?

joeparsons:
Maybe not.  I suppose `dist` may have already been (manually) updated recently enough.

kevcooper:
there are some new files that get created if I delete `dist` first though

kevcooper:
never mind, they're not really new they just have the wrong name and are in the wrong location

PR #89, opened 2020-06-17
Use an ephemeral local Docker image preconfigured with npm — Closes #73

### LINKED ISSUE #73: The Docker-based build environment for CI and local development is unstable.
## What is the problem that we want to solve?
The general-purpose Docker image used in the builds needs several additional steps that are slow and can be unreliable, particularly for `npm install`. We should investigate whether derived temporary Docker containers work around this for both local development and GitHub Actions, or whether GitHub Actions can handle the CI aspect, with more work being needed for local development.

## Conditions of satisfaction.         
- [ ] *Either* a proof-of-concept improved Docker-based environment for CI and local dev.
- [ ] *Or* a GitHub Actions-specific improvement proof-of-concept.
- [ ] An estimate of time and effort is provided for the recommendation.

mmunro-ltrr:


.github/workflows/release.yml
```diff
@@ -20,6 +20,7 @@ on:
   push:
     branches:
       - master
+      - user-story/60-docs
```
kevcooper, line 23: Remove this

kevcooper:
I've been testing this and two things have come up so far:
- The docker layers don't seem to cache and the "Build Variables" step seems to take ~6 minutes every time this action runs
- Having issues syncing the _site/ directory, is this still where the artifact ends up?

kevcooper:
another potential issue: "build Variables" also runs even during the "back to dev" commits... we might want to consider the docker build as a separate step so we can add `if: ${{ endsWith(env.AZ_VERSION, '-dev') == false }}` to it

kevcooper:
closes #95

PR #96, opened 2020-06-23
WIP #93: Add JS bundling/compiling tools and configs back to Arizona Bootstrap

joeparsons:
WIP branch to hopefully fix #93.

- Reverted UADIGITAL-2058 changes
- Copied latest configs/commands/dependencies from twbs/boostrap
- Added hacky offcanvas JS module/plugin

PR #108, opened 2020-08-07
Fixes #93: Restored JS tooling & configs back to AZ Bootstrap.

### LINKED ISSUE #93: Add JS tooling back to Arizona Bootstrap
## Motivation
We've decided to re-vamp our mobile navigation (see #82, #90).  In order to support the navigation style/functionality we've decided on, we will need to add a small amount of custom JS to arizona bootstrap however 

About a year ago we decided to simplify our existing package.json JS tasks for UADIGITAL-2058 which removed many of the tools/configuration needed to support custom JS in arizona-bootstrap so we'll need to look at adding that back.

Add back JS compiling/bundling tooling and configuration (copied from twbs/bootstrap) to arizona-bootstrap.

joeparsons:
This is a "clean" version of #96 rebased on master without any of the navigation menu stuff.

The NPM packages being added here are not in sync with the versions in the latest twbs 4.x release anymore so that probably still needs work.


mmunro-ltrr:
Tried testing locally: the full JS build might not be happening: I'll have to look in more detail.

joeparsons:
I've updated this with the latest dependency versions & config files from twbs v4.5.2

mmunro-ltrr:
In a few places the way the hash that's used as a tag for the Docker image is computed has changed. The intent seems to be to summarize all changes to the contents of the `scripts` subdirectory by combining them in a tarfile & taking the hash of that, with the assumption that the hash will not change if none of the files within the subdirectory have changed. The Docker image does depend on the scripts in addition to the lockfiles that are currently used in computing the final hash, so that seems like a good thing. Unfortunately the tarfile hash is only consistent within a particular clone of the repository: the timestamp on the `scripts` directory will change on different copies, this change will propagate to the tarfile contents, and the hashes won't be the same, even with identical files within the subdirectory. The `scripts/build-container.sh` script is still using the older hash computation.

joeparsons:
> In a few places the way the hash that's used as a tag for the Docker image is computed has changed. The intent seems to be to summarize all changes to the contents of the `scripts` subdirectory by combining them in a tarfile & taking the hash of that, with the assumption that the hash will not change if none of the files within the subdirectory have changed. The Docker image does depend on the scripts in addition to the lockfiles that are currently used in computing the final hash, so that seems like a good thing. Unfortunately the tarfile hash is only consistent within a particular clone of the repository: the timestamp on the `scripts` directory will change on different copies, this change will propagate to the tarfile contents, and the hashes won't be the same, even with identical files within the subdirectory. The `scripts/build-container.sh` script is still using the older hash computation.
> 
> But the same approach used with the lockfiles should work to summarize the directory, something like
> 
> ```
> find scripts -type f -exec sha256sum {} \;  | awk '{ print $1 }' | sha256sum - | awk '{ print $1 }'
> ```

Thanks @mmunro-ltrr!  I meant to ask you if there was a better way to do that.  I'll make this change.

mmunro-ltrr:
Unfortunately Shellcheck doesn't like the use of `sha256sum` with `find` in scripts/buildcontainer.sh, because there we redefine `sha256sum` to be whatever the native SHA256 utility is called (errors https://www.shellcheck.net/wiki/SC2032 and https://www.shellcheck.net/wiki/SC2033). It doesn't matter in GitHub actions, because there the name is already the native one. Looking at a better way of doing this....

mmunro-ltrr:
Just tried to split off synchronizing with the latest v4.5.2 twbs upstream release as a distinct PR.

mmunro-ltrr:
Splitting off the SHA256 hash improvements as a distinct issue would make sense as well (a needed improvement even without restoring the JS tooling).

PR #190, opened 2020-11-04
Feature/182 add arizona header into arizona bootstrap

### LINKED ISSUE #182: Add arizona header into arizona bootstrap.
## Motivation
We need the style for the Arizona Header across our web ecosystem, not just is arizona quickstart.

If creating a non Quickstart site i have to recreate the header.

Due to the recent addition of the off-canvas menu, now would be a good time to add this code because we can not use the traditional container grid based design for the header.

Alter the menu and search icons to not appear where they do, and instead appear within a container.

trackleft:
Closes #182  Because I need it for the off-canvas navbar.
If anyone wants to test on small phones, please do.
Or you could use xcode


camikazegreen:
Link to the review site: https://review.digital.arizona.edu/arizona-bootstrap/feature/182-add-arizona-header-into-arizona-bootstrap/docs/2.0/components/arizona-header/

site/docs/2.0/components/arizona-header.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Wordmark header
+The header is to be used on all subdomains of arizona.edu. Extending the header
+to save room on your site, especially in mobile, with nav and search buttons is
+supported as shown below.
+
+### This header should only be used when near a Block "A" logo.
+
+<header class="bg-red arizona-header" id="header_arizona">
+  <div class="container l-container">
```
joeparsons, line 17: Is `l-container` an actual class that is defined anywhere?

danahertzberg:
Please match mockup styles: https://xd.adobe.com/view/076fa3a3-f5ea-4060-8ab1-31700895aeed-523c/specs/

camikazegreen (approved):

trackleft:
Hoping to get this merged soon so we can review the quickstart off canvas nav ticket with all the styles in place.

https://github.com/az-digital/az_quickstart/pull/335

Review here: https://review.digital.arizona.edu/arizona-bootstrap/feature/182-add-arizona-header-into-arizona-bootstrap/docs/2.0/components/arizona-header/

scss/_custom-styles.scss
```diff
@@ -132,3 +132,64 @@
...
+}
+.block-a-line-logo {
+  width: auto;
+  height: 36px;
+  margin: 5px 20px 0 10px;
+}
+.arizona-line-logo {
+  width: 211px;
+  height: 16px;
+  margin: 17px 20px 17px 10px;
```
joeparsons, line 149: Is there a reason the left and right margin values are different?  I might be missing something obvious but it seems to me that it'd be better to have the left margin be 20px as well (?).

danahertzberg:
@trackleft Please have the buttons be 60x50 px in size without space in-between.

scss/_custom-styles.scss
danahertzberg: I agree, the left margin should also be 20px to give the logo even spacing.

danahertzberg:
Also, it looks like the logo stops shrinking on small screens
Screen Shot 2020-11-12 at 12 45 43 PM

trackleft:
While I don't disagree with the points made on the image being scaled and the margin around the image needing to be sorted out, but I wonder if this should be part of a follow up ticket or two, since those two things have contextual implications with the grid and the logos beneath them that might be better handled when we have sites with the logos on them.  Additionally, the logo seems to get somewhat blurry when it is scaled down programmatically.

I'll make the buttons 60px wide, I wrongly assumed they were squares and didn't look at the width attribute when I set them at 50px.

scss/_custom-styles.scss
```diff
@@ -132,3 +132,64 @@
...
+  font-size: 24px;
+  line-height: 24px;
+}
+.btn.btn-redbar {
+  width: 60px;
+  height: 50px;
+  padding: 0;
+  color: #fff;
+  background-color: #ab0520;
+  border: 2px solid #ab0520;
```
danahertzberg, line 187: Not sure why a border was added. I don't think it's necessary, but we can clean it up later.

danahertzberg:
> While I don't disagree with the points made on the image being scaled and the margin around the image needing to be sorted out, but I wonder if this should be part of a follow up ticket or two, since those two things have contextual implications with the grid and the logos beneath them that might be better handled when we have sites with the logos on them. Additionally, the logo seems to get somewhat blurry when it is scaled down programmatically.
> 
> I'll make the buttons 60px wide, I wrongly assumed they were squares and didn't look at the width attribute when I set them at 50px.

Thanks for updating the button widths. For the logo, would it be appropriate for this PR to not have the logo and just focus on providing a red bar with buttons? Then the logo is added in Quickstart? Not sure how this plays with brand standards. Just some thoughts.

barrettbaffert (approved):
I agree that some follow up tickets might be necessary. But I want to see it in the quickstart context first and don't think there's harm in merging this now

camikazegreen (approved):

mmunro-ltrr (approved):
Checked locally.

PR #202, opened 2020-12-02
update version to 2.0.5

trackleft:


joeparsons:
This is no longer needed (a separate PR was merged instead)

PR #224, opened 2021-02-23
Update Link description

its-me-mario:
Removed "on hover" to more clearly indicate that the default state of links should be underlined.

For reference:
https://webaim.org/techniques/hypertext/link_text#text
"The link text must have a 3:1 contrast ratio from the surrounding non-link text." - WCAG 2.0 Level A
Current contrast ratio between text and link color is 1.17:1, so removing default underlines from links would not be sufficient differentiation.

### DIFF
--- a/site/docs/2.0/content/typography.md
+++ b/site/docs/2.0/content/typography.md
@@ -165,7 +165,7 @@ While not shown above, feel free to use `<b>` and `<i>` in HTML5. `<b>` is meant
 
 ## Links
 
-Default links include a red text color and an underline on hover. You can change a link's default styling by adding additional classes to it (e.g., `.btn`, `.btn-red`, etc.). Links inside navbars, breadcrumbs, and other components receive their own default styling. 
+Default links include a red text color and an underline. You can change a link's default styling by adding additional classes to it (e.g., `.btn`, `.btn-red`, etc.). Links inside navbars, breadcrumbs, and other components receive their own default styling. 
 
 {% capture example %}
 <a href="#">Default Link</a>


trackleft:
Talked to Thomas McCarthy at admissions and he said:
```
I think they should remain the same color as the text but be underlined and semibold
then on the hover they become bold and red
```
So, that is one other designer who is good with this change, which I am all for as well, though I think we need a code change as well.

trackleft:
FYI ua-bootstrap looks like this: http://uadigital.arizona.edu/ua-bootstrap/
default links


arizona-bootstrap looks like this https://digital.arizona.edu/arizona-bootstrap/docs/2.0/content/typography/#links
default-links-bs4

its-me-mario:
I'm okay with those changes too, assuming that there isn't bold underlined text mixed with links.

It might be a good idea to mention that link color should be consistent with the surrounding text color `color: currentColor` instead of saying a color.

danahertzberg:
I would advocate for keeping the color with the knowledge that the color is not what makes us accessible. I think it's a nice added visual.

Suggested text change: 
- Was: Default links include a red text color and an underline on hover.
- Proposed: Default links are chili red, bold, and underlined. 

Question for @its-me-mario: Are there accessibility guidelines about hover effects on links?

its-me-mario:
@danahertzberg As far as I can see, it just wants there to be some sort of visual change when you hover and focus on a link. e.g. changes to background-color, color, text-decoration

https://webaim.org/techniques/hypertext/link_text#hover

trackleft:
https://webaim.org/articles/contrast/

> This becomes increasingly difficult if the links change color on hover or focus each of the colors in these states must also have at least a 4.5:1 contrast ratio difference with the background. WCAG 2.0 and Link Colors on the WebAIM blog explores the WCAG requirements for link colors in more depth.



https://webaim.org/blog/wcag-2-0-and-link-colors/

> Link states
> This becomes more complicated if you want to provide various link states (hover/focus, visited, active) colors. If you go with the traditional blue, red, and purple colors for link states, each of these three colors would have to fit into the narrow window of available contrasts.
> 
> Some testing shows that this is possible… barely. The following colors for link states meet these requirements for a black on white page:
> 
> Normal: #3344dd
> Hover/Focus/Active: #bb1122
> Visited: #884488
> 
> But there’s not much wiggle room – if you change any of these colors much you’ll break either the 3:1 requirement to black or the 4.5:1 requirement to white. The W3C does provide a list of web-safe colors that fit within these contrast requirements. But if your page doesn’t have black on white (or vice versa), it quickly becomes impossible to get any one color to work, let alone two or three colors for the various link states.
>

trackleft:
Note: Obvious other things that contribute to the accessibility/readability of text on screen.
`font-weight`
`font-size`
`background-color`

its-me-mario:
Just wanted to point out, in case I didn't understand the Webaim docs...

The 3:1 contrast requirement is only if you don't underline links.
https://webaim.org/articles/contrast/
> For optimal accessibility and usability, maintain the underline on links.  Otherwise two conditions must be met:

https://webaim.org/blog/wcag-2-0-and-link-colors/
> This means that if you do not underline your links (or provide some other non-color designator for links), that the links must be sufficiently different in contrast from the surrounding text.

trackleft (approved):

danahertzberg (approved):

PR #225, opened 2021-02-24
Update release action to use manual trigger

### LINKED ISSUE #137: Release Automation Improvements
Now that `workflow_dispatch` has a web UI, we'll want to add a new action to allow us to manually initiate a workflow that creates the release branch, builds the project, and creates a PR for the new version.

kevcooper:
The release action will now use a `workflow_dispatch` trigger with the new version number as input. It will then run a build, commit and push the results, and tag a release.

Closes #137


mmunro-ltrr:
Just checked to see if the recent changes have now broken this, & it looks like they have. I'll attempt a repair.

mmunro-ltrr:
A suggested update to account for recent changes is at the branch https://github.com/az-digital/arizona-bootstrap/tree/issue/137plusfixes

mmunro-ltrr:
Further thoughts on the release process: we shouldn't look for a pre-existing Docker image, but unconditionally build a new one, containing an npm environment based on a clean `npm install` & the current package.json file. The small risk is that this will include some update (& an updated package-lock.json file) that none of the prior review & testing had yet seen, but this is less than the risk of picking up some old & possibly outdated configuration fixed within an existing image. The other branch doesn't yet do this.

mmunro-ltrr:
Modified the `issue/137plusfixes` branch to drop the search for a pre-existing image & build a new one unconditionally.

mmunro-ltrr:
# Create Release (comments on the Docker steps)
## Checkout Code
Start in a Git clone of the repository, with the current state of the `main` branch checked out.
## Set Variables for Docker Images
Make a hash from various files: unlike some of the other GitHub Actions in this repository, where the hash is a key used to find any pre-built Docker image to use, a release only uses this to name a temporary container.
```
oldhash=${{ hashFiles('Dockerfile', 'package.json', 'package-lock.json', 'scripts/*') }}
```
Set up with the deprecated GitHub Packages Docker registry: this domain is changing to `ghcr.io` for the recommended GitHub Container Registry, and stored images there won't be tied to a specific repository.
```
registry='docker.pkg.github.com'
imageprefix="${registry}/${GITHUB_REPOSITORY}/"
```
The image name will be the same one used in other actions, but this is arbitrary (it could be anything).
```
imagename='az-nodejs-ephemeral'
imagestem="${imageprefix}${imagename}:"
```
Make these variables available to subsequent actions, not just here.
```
echo "AZ_DOCKER_REGISTRY=${registry}" >> ${GITHUB_ENV}
echo "AZ_OLD_HASH=${oldhash}" >> ${GITHUB_ENV}
echo "AZ_IMAGE_STEM=${imagestem}" >> ${GITHUB_ENV}
```
Put the default directory paths within the container into variables that subsequent actions will see in the environment; these are lifted from the Dockerfile, and although an image build could override the defaults, it would need explicit build arguments (just re-setting the variables here would not suffice). The source directory is the one usually shared between the container and the host system, the frozen directory holds a fully built-out npm setup (including `package.json`, `package-lock.json`, and `node_modules`) which is baked into the Docker image and internal to the container.
```
echo "AZ_BOOTSTRAP_SOURCE_DIR=arizona-bootstrap-src" >> ${GITHUB_ENV}
echo "AZ_BOOTSTRAP_FROZEN_DIR=/azbuild/arizona-bootstrap" >> ${GITHUB_ENV}
```
## Docker Authentication
Log in to the Docker repository once for all subsequent steps: this is for the deprecated package repository, and code examples for the new one are showing something like `run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin` instead of what appears here.
```
docker login "$AZ_DOCKER_REGISTRY" -u "$GITHUB_ACTOR" -p ${{ secrets.GITHUB_TOKEN }}
```
## Unconditionally Rebuild and Save the Docker Image
Build an image using the Dockerfile in the current repository checkout, suppressing all messages apart from the final output of the image ID (as a sha256 hash); the build will include a full npm install setup, using the current package.json file. Save the ID in a local environment variable.
```
workingtitle=$(docker build -q . )
```
Make an arbitrary semi-unique name (the content of the hash doesn't matter, and it might even be better to use something like a timestamp instead).
```
tempname="old${AZ_OLD_HASH}"
```
Spin up but immediately shut down a container, using the newly created image and giving it the arbitrary name. Note that this does not share any directories with the host.
```
docker run --name "$tempname" "$workingtitle" true
```
Using a host-based Docker command, extract the container's pre-built npm setup (actually fixed as part of the image), and dump it into the current directory, overwriting anything that's already present.
```
docker cp "${tempname}:${AZ_BOOTSTRAP_FROZEN_DIR}/." .
```
The container was simply a means to extract the npm setup, and is not useful for anything else, so can be deleted immediately.
```
docker rm "$tempname"
```
Make a tag for the image which depends on the contents of the current files, including some that have just come out of the extracted npm setup, and save the tagged image. Note that (unlike some of the other GitHub Actions in this repository) nothing else will probably use the saved and tagged image, so a refactoring that drops the authentication & other uses of the repository would be feasible.
```
lockhash=${{ hashFiles('Dockerfile', 'package.json', 'package-lock.json', 'scripts/*') }}
ephemeral="${AZ_IMAGE_STEM}${lockhash}"
docker tag "$workingtitle" "$ephemeral"
docker push "$ephemeral"
```
Make the full image name available to subsequent actions.
```
echo "AZ_EPHEMERAL_IMAGE=${ephemeral}" >> ${GITHUB_ENV}
```
## Build Variables
Put the manually supplied next version number in the environment subsequent actions will see.
```
echo "AZ_VERSION=${{ github.event.inputs.version }}" >> ${GITHUB_ENV}
```
## Update Version
Create the mandatory settings for pushing Git changes. Docker containers will not be able to use these directly, so some of the documented examples for having the `npm version` command immediately make Git-based updates will not be workable here; only the host, not the container, can do this.
```
git config --global user.email "${GITHUB_ACTOR}@users.noreply.github.com"
git config --global user.name "${GITHUB_ACTOR}"
```
At this point the current working directory contains the files from the main branch, with a freshly updated npm setup copied on top. There's a new Docker image containing an identical npm setup, and the supporting environment (the specific version of Node.js used to create it, and so on). Start a Docker container using the image, this time sharing the current directory with the container, providing the new version number for the release, and specifying that the `scripts/create-release.sh` script should run within the container. This will perform the following sequence of operations:
  * Re-synchronize the saved npm setup with the working directory (this is almost certainly redundant, and might be dropped).
  * Explicitly invoke the `npm version --unsafe-perm --no-git-tag-version ${AZ_RELEASE_VERSION}` command, updating the version numbers in both the package.json and package-lock.json files; because those files are in the working directory (not the previously saved frozen copies), and the container & host share this directory, the change should immediately propagate to the host. *I can't remember why the --unsafe-perm flag was necessary.*
  * Implicitly re-build the CSS and JS files in the `dist` directory subtree, because the package.json file's `scripts:` section includes a line `"version": "npm run dist",` and the `npm version` command will run this automatically; again, the directory is shared with the host, so the change should appear outside the container.
```
docker run --rm -e "AZ_RELEASE_VERSION=${AZ_VERSION}" -v $(pwd):"/${AZ_BOOTSTRAP_SOURCE_DIR}" "$AZ_EPHEMERAL_IMAGE" create-release
```
The container should stop after the version number bump and dist subdirectory update, and normal Git operations should update the repository. *Does this work in practice, since if the push would be going directly to the main branch? Also, we've saved the AZ_VERSION variable, but seem to have stopped using it here.*
```
git add dist package.json package-lock.json
git commit -m '${{ github.event.inputs.version }}'
git push
```
Finally, save the raw SHA code of the commit to a variable. *A subsequent step might duplicate this.*
```
echo "RELEASE_SHA=$(git rev-parse HEAD)" >> ${GITHUB_ENV}
```

joeparsons (approved):

mmunro-ltrr (approved):

Issue #213, opened 2021-01-08
Update Arizona header once off-canvas markup is finalized.


trackleft:
Update [Arizona Header](https://github.com/az-digital/arizona-bootstrap/blob/main/site/docs/2.0/components/arizona-header.md) with new search and menu open buttons for mobile once they are finalized. Use the same markup for both.

_Originally posted by @trackleft in https://github.com/az-digital/arizona-bootstrap/pull/209#discussion_r554201634_

**Requirements:**
- Markup on AZ Barrio should match Arizona Bootstrap (reduce extra markup as much as possible)
    - Update main menu template
- Address scrolling issues on menu (should scroll) and background (should not scroll https://github.com/az-digital/az_quickstart/issues/614)

**Potential future ticket:**
- Implement menu with modal component instead of collapse component (see BS5)

akslay:
@trackleft Do you know if this has already been addressed/resolved by other pull requests (i.e., #566) ? The markup for the offcanvas nav and Arizona Header appear to be the same as what we have in Bootstrap.

> Markup on AZ Barrio should match Arizona Bootstrap (reduce extra markup as much as possible. Update main menu template

akslay:
Did a little bit of investigation, and it looks like the Arizona Header tasks are already complete (markup matches). On the Drupal side, however, you still cannot scroll when the mobile nav extends past the side of the screen fold. This is not an issue in AZ Bootstrap (works fine in the documentation) so a separate issue needs to be opened/fixed on the AZ Quickstart side.

The issue of resolving background scroll while you are scrolling through the mobile nav should be a separate issue as well, since it's unrelated to the Arizona Header markup and would be an independent improvement to the off-canvas mobile nav. Not sure if we want to "morph" this ticket into addressing that scrolling issue, or if we want to create a fresh issue.

danahertzberg:
@akslay Let's keep this ticket for the menu scrolling issue. I will create a separate issue for the GA release for the background scroll issue.

akslay:
> Let's keep this ticket for the menu scrolling issue. I will create a separate issue for the GA release for the background scroll issue.

I think we actually need a separate issue for the menu scrolling not working, because the issue isn't in Arizona Bootstrap, it's in AZ Quickstart. So the fix would be in the CSS of the `az_quickstart` repo rather than the `arizona_bootstrap` repo. Did you want to keep this ticket for the background scroll instead?

danahertzberg:
[...]

PR #240, opened 2021-03-09
Closes #223: Fixes for various off-canvas navigation bugs.

### LINKED ISSUE #223: Main navigation font weight, whitespace, and scrolling bugs
## Problem/Motivation
Larger screen size main navigation changed to be inaccessible after the off-canvas PR was merged and it was found that mobile navigation is not scrolling.

### Description and proposed solutions (screenshots further below)
Before: https://dev-sasg-registrar.pantheonsite.io
After: http://azqs-sasg-registrar.pantheonsite.io

#### Font weight:
"Font weight for the added off-canvas nav is impacting the desktop breakpoint and making it very difficult to read. Font weight change needs to be moved to a mobile media query." - @akslay 
Specifically this CSS rule: `.navbar-offcanvas .dropdown-menu .dropdown-item`

#### Whitespace:
Whitespace property was changed to have text wrap. This declaration will need to be added to a media query.

#### Scrolling:
Mobile nav is not scrolling (unsure of fix).

### To Reproduce
Create a navigation menu with several items and dropdowns.

##  Screenshots
reg_update_before
reg_update_after

joeparsons:
Made changes suggested by @akslay to fix off-canvas navigation bugs seen on in-devlopment Registrar website.

### DIFF
--- a/scss/_navbar-offcanvas.scss
+++ b/scss/_navbar-offcanvas.scss
@@ -81,10 +81,6 @@
     }
   }
 
-  .nav {
-    max-height: 85vh;
-    flex-wrap: initial;
-  }
 
   .navbar-nav {
     width: 100%;
--- a/scss/_navbar-offcanvas.scss
+++ b/scss/_navbar-offcanvas.scss
@@ -81,12 +81,14 @@
     }
   }
 
+  .nav {
+    max-height: 85vh;
+    flex-wrap: initial;
+  }
+
   .navbar-nav {
     width: 100%;
     overflow-y: auto;
-    .nav-link {
-      font-weight: 500;
-    }
     .show > .nav-link {
       color: $navbar-offcanvas-link-color;
       background-color: $navbar-offcanvas-bg-expanded;
@@ -103,7 +105,7 @@
   }
 
   .nav-item {
-    
+
     &.active, {
       .nav-link {
         color: $navbar-offcanvas-link-color;
@@ -156,7 +158,7 @@
 
     .dropdown-item {
       padding: $navbar-offcanvas-dropdown-link-padding;
-      font-weight: 200;
+      font-weight: 400;
       white-space: normal;
     }
   }
@@ -225,6 +227,7 @@
     align-items: center;
     -ms-flex-pack: justify;
     justify-content: space-between;
+    max-width: 100%;
     padding: 0 1rem;
     padding: initial;
     overflow-y: initial;
@@ -252,6 +255,9 @@
     }
     .navbar-offcanvas-header {
       display: none;
+    }
+    .nav {
+      flex-wrap: wrap;
     }
     .navbar-nav {
       display: flex;
@@ -333,6 +339,7 @@
       position: absolute;
       .dropdown-item {
         padding: $dropdown-item-padding-y $dropdown-item-padding-x;
+        white-space: nowrap;
       }
     }
     .dropdown-toggle {
@@ -348,6 +355,7 @@
         border-right: 0.3em solid transparent;
         border-bottom: 0;
         border-left: 0.3em solid transparent;
+        margin-left: 0.225em
       }
     }
   }
--- a/scss/_navbar-offcanvas.scss
+++ b/scss/_navbar-offcanvas.scss
@@ -251,9 +251,6 @@
     .navbar-offcanvas-header {
       display: none;
     }
-    .nav {
-      flex-wrap: wrap;
-    }
     .navbar-nav {
       display: flex;
       flex-direction: column; // cannot use `inherit` to get the `.navbar`s value
--- a/scss/_navbar-offcanvas.scss
+++ b/scss/_navbar-offcanvas.scss
@@ -84,6 +84,9 @@
   .navbar-nav {
     width: 100%;
     overflow-y: auto;
+    .nav-link {
+      font-weight: 500;
+    }
     .show > .nav-link {
       color: $navbar-offcanvas-link-color;
       background-color: $navbar-offcanvas-bg-expanded;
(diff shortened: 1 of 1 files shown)


mmunro-ltrr (dismissed):
Quick test on a local build: I'm not sure I fully exercised the bugs, but it would be good to merge in advance of other changes.

geijutsuka (dismissed):
Looks like the desktop styles are back to normal (see before and after below). On a very small screen, the menu is long enough to demonstrate that the scrolling is working as well.
Before:
Screen Shot 2021-03-09 at 10 21 11 AM

After:
Screen Shot 2021-03-09 at 10 21 01 AM

scss/_navbar-offcanvas.scss
```diff
@@ -333,6 +339,7 @@
       position: absolute;
       .dropdown-item {
         padding: $dropdown-item-padding-y $dropdown-item-padding-x;
+        white-space: nowrap;
```
danahertzberg, line 342: This causes the menu to get cut off with long menu titles. Looking into a solution now (we also had to fix this in QS1) but wanted to ensure it was shared before this got merged.

Quickest solution would be to not include this until a solution for all screen sizes is discovered. Not including it just makes the dropdown menu skinnier than desired for now.

image

scss/_navbar-offcanvas.scss
```diff
@@ -253,6 +256,9 @@
     .navbar-offcanvas-header {
       display: none;
     }
+    .nav {
+      flex-wrap: wrap;
+    }
```
danahertzberg, line 261: Where is this class being used? I'm not seeing it in the review site.

scss/_navbar-offcanvas.scss
```diff
@@ -156,7 +158,7 @@
 
     .dropdown-item {
       padding: $navbar-offcanvas-dropdown-link-padding;
-      font-weight: 200;
+      font-weight: 400;
```
danahertzberg, line 161:
```suggestion
```
Recommend removing since it's being set to the same value as the default .dropdown-item class.

scss/_navbar-offcanvas.scss
```diff
@@ -81,12 +81,14 @@
...
+    max-height: 85vh;
+    flex-wrap: initial;
+  }
+
   .navbar-nav {
     width: 100%;
     overflow-y: auto;
-    .nav-link {
-      font-weight: 500;
-    }
```
danahertzberg, line 89: Keep this and add `.nav-link {font-weight: 700;}` to the media-breakpoint-up(lg) query. Then the off-canvas styles won't change from how they were before. 

I would add this as suggestions, but GitHub isn't allowing me. :(

scss/_navbar-offcanvas.scss
```diff
@@ -81,12 +81,14 @@
     }
   }
 
+  .nav {
+    max-height: 85vh;
+    flex-wrap: initial;
+  }
```
danahertzberg, line 87: Is this to address the overflow scroll issue? Could you share steps to reproduce the scrolling issue? And where is this class used within the `navbar-offcanvas` div (I think I'm missing something)?

scss/_navbar-offcanvas.scss
danahertzberg, line 89: Keep this and add .nav-link {font-weight: 700;} to the media-breakpoint-up(lg) query. Then the off-canvas styles won't change from how they were before.

I would add this as suggestions, but GitHub isn't allowing me. :(

danahertzberg (changes requested):

scss/_navbar-offcanvas.scss
joeparsons: In AZ Barrio, the `<ul>` rendered in the main menu template has this class applied to it (see [themes/custom/az_barrio/templates/navigation/menu--main.html.twig#L33](https://github.com/az-digital/az_quickstart/blob/main/themes/custom/az_barrio/templates/navigation/menu--main.html.twig#L33)) and that template is used to render the main menu inside the `navigation_offcanvas` region in the page template when the `Use Navbar Off Canvas` theme setting is enabled (see [themes/custom/az_barrio/templates/page.html.twig#L138-L161](https://github.com/az-digital/az_quickstart/blob/main/themes/custom/az_barrio/templates/page.html.twig#L138-L161))

danahertzberg: After discussing with @joeparsons, he is going to create a separate PR to remove the `nav` class from `navbar-nav` since they should not be used together.

danahertzberg: Remove for now, address Barrio/Bootstrap discrepancy issue and see if there are scrolling issues with this ticket. https://github.com/az-digital/arizona-bootstrap/issues/213

danahertzberg:
```
  .nav {
    max-height: 85vh;
    flex-wrap: initial;
  }

Comment (danahertzberg) [inline scss/_navbar-offcanvas.scss:87]:
```suggestion

```

scss/_navbar-offcanvas.scss
```diff
@@ -81,12 +81,14 @@
...
+
```
danahertzberg, line 88:
```suggestion
```

danahertzberg (changes requested):

scss/_navbar-offcanvas.scss
danahertzberg, line 261:
```suggestion
```

danahertzberg (changes requested):

danahertzberg:
Next steps: https://github.com/az-digital/arizona-bootstrap/issues/213

danahertzberg (approved):

mmunro-ltrr (approved):

PR #310, opened 2021-06-17
Closes #243: Adding class to auto hyphenate long words

### LINKED ISSUE #243: Card titles break word
_SASG found original bug_

Longer card titles wrap instead of breaking the whole word to the next line.

smayden:
This PR creates a custom of class .text-hyphen to set hyphens: auto. 

SASG reported a problem with words breaking without hyphens in #243. Dana and I were not able to recreate their exact issue, but did not that exceptionally long words can break this way if they are longer than the column size (particularly when there are four cards on a page with sidebar nav).

The .text-hyphen class can be used sparingly in these cases to allow for auto hyphenation of singe words. The documentation for Text Utilities has been updated with an example and also the caution that this class should not be used on paragraphs or longer spans of text.

Review site: https://review.digital.arizona.edu/arizona-bootstrap/issue/243/docs/2.0/utilities/text/

### DIFF
--- a/scss/_custom-styles.scss
+++ b/scss/_custom-styles.scss
@@ -108,6 +108,12 @@
   font-family: $font-family-sans-serif;
   font-weight: 500;
 }
+
+// >> Text Hyphenation.
+.text-hyphen {
+  hyphens: auto;}
+}
+
 // >> Lists
 .ul-triangles {
   overflow: hidden;
--- a/site/content/docs/2.0/utilities/text.md
+++ b/site/content/docs/2.0/utilities/text.md
@@ -71,10 +71,10 @@ Prevent long strings of text from breaking your components' layout by using `.te
 
 ## HYPHENATE WORD
 
-Allow long strings of text to break to the next line with a hyphen automatically placed at the break point by using `.text-hyphen` to set `hyphens: auto`.
+Allow long strings of text to break to the next line with a hyphen automatically placed at the break point by using `.text-hyphen` to set `hyphens: auto`. This can be applied as a span on single words that are longer than the column width.
 
 {{< example >}}
-<div class="card col-1">
+<div class="card col-2">
 	<div class="card-body">
 		<p>We should all learn about <span class="text-hyphen">Tranformationaltotality</span></p>
 	</div>
--- a/site/content/docs/2.0/utilities/text.md
+++ b/site/content/docs/2.0/utilities/text.md
@@ -74,7 +74,11 @@ Prevent long strings of text from breaking your components' layout by using `.te
 Allow long strings of text to break to the next line with a hyphen automatically placed at the break point by using `.text-hyphen` to set `hyphens: auto`.
 
 {{< example >}}
-<p class="text-hyphen">mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm</p>
+<div class="card col-1">
+	<div class="card-body">
+		<p>We should all learn about <span class="text-hyphen">Tranformationaltotality</span></p>
+	</div>
+</div>
 {{< /example >}}
 
 ## Text Transform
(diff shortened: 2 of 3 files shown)


mmunro-ltrr:
Since hyphenation makes continuous text look much better, it seemed strange that Bootstrap didn't include it already. The last explicit use of it was a deprecated vendor mixin back in v3 (where it was just a workaround to help browsers that didn't support the `hyphens` property in CSS), there's a documentation note about this in v4, and not a whisper in v5. Hyphenation depends on having the (human) language of the text specified somewhere, and on the browser having the appropriate hyphenation dictionary,
https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens
so perhaps this was all too fragile for Twitter Bootstrap to support it.  The Arizona Bootstrap documents do include an explicit `lang="en"` specification, so it should work there, but perhaps the documentation should include a warning about using it in arbitrary contexts.

danahertzberg:
Thanks @mmunro-ltrr! I totally understand how machine-generated hyphenation can be problematic in the human language. I agree we should explicitly state the intended use and potential pitfalls of using this class arbitrarily to the bootstrap documentation.

This class is intended to be used manually on specific words where an exception is needed. For example, if the word "transformation" extends beyond its boundaries on some screen sizes but not others, we could add this new class in a span around it to hyphenate only when necessary. Applying the hyphens: auto; to all text is problematic since it hyphenates all words regardless of how they display and could fit on the next line.

scss/_custom-styles.scss
```diff
@@ -108,6 +108,12 @@
   font-family: $font-family-sans-serif;
   font-weight: 500;
 }
+
+// >> Text Hyphenation.
+.text-hyphen {
+  hyphens: auto;}
```
trackleft, line 114:
```suggestion
  hyphens: auto;
```

smayden:
Thanks for your input @mmunro-ltrr and @danahertzberg (and also @trackleft). This is now working and I've updated the documentation. Ready for review.

Review site: https://review.digital.arizona.edu/arizona-bootstrap/issue/243/docs/2.0/utilities/text/

site/content/docs/2.0/utilities/text.md
```diff
@@ -69,6 +69,18 @@ Prevent long strings of text from breaking your components' layout by using `.te
 <p class="text-break">mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm</p>
 {{< /example >}}
 
+## HYPHENATE WORD
```
mmunro-ltrr, line 72: One minor cosmetic gripe: none of the other h2 headings in this part of the docs are all caps.

mmunro-ltrr (dismissed):

site/content/docs/2.0/utilities/text.md
smayden: Great catch! Thanks

mmunro-ltrr (approved):

site/content/docs/2.0/utilities/text.md
```diff
@@ -69,6 +69,18 @@ Prevent long strings of text from breaking your components' layout by using `.te
...
+## Hyphenate Word
+
+Allow long strings of text to break to the next line with a hyphen automatically placed at the break point by using `.text-hyphen` to set `hyphens: auto`. This should only be applied as a span on a single word that may be longer than the column width at some screen sizes. Do not use on paragraphs or longer spans of text.
+
+{{< example >}}
+<div class="card col-2">
+	<div class="card-body">
+		<p>We should all learn about <span class="text-hyphen">Tranformationaltotality</span></p>
+	</div>
+</div>
```
danahertzberg, line 81: Nitpicking some formatting. Also, nice long word!
```suggestion
<div class="card col-2">
  <div class="card-body">
    <p>We should all learn about <span class="text-hyphen">Tranformationaltotality</span>.</p>
  </div>
</div>
```

smayden:
Updated with Dana's suggestions

mmunro-ltrr (approved):

danahertzberg (approved):

joeparsons (approved):

PR #313, opened 2021-06-18
Closes #294: Adding border-thick class and documentation

### LINKED ISSUE #294: Add new .border-thick and border color classes
## Motivation
I want to recreate the style of the "Extra Info" paragraph type.

## Context
This is currently created by adding these classes to the surrounding div
```
card card-block border-top-accent-azurite mb-0
```
`.border-top-accent-azurite` currently uses a 4px border width. I'm proposing to use 0.25rem border width to match the callout style.

## Resolution
Add this new class and CSS
```
.border-thick {
    border-width: .25rem!important;
}
```
Remove importance from the border class to allow brand color border classes to be used (ie. border-red)
```
.border, .panel {
    border: 1px solid #dee2e6!important;
}
```
Update Arizona Bootstrap documentation
- Add examples: https://digital.arizona.edu/arizona-bootstrap/docs/2.0/utilities/borders
- Migration of `border-top-accent-{azurite,oasis}` classes: https://digital.arizona.edu/arizona-bootstrap/docs/2.0/migration/

danahertzberg:
Closes #294 

Updates include
- Added `!important` to border-color classes to allow border-color classes to work as expected
- Added new `.border-thick` class to allow for thicker border where needed (matches thickness of callout border)
- Updated documentation with example
- Updated documentation about migration

Review site: https://review.digital.arizona.edu/arizona-bootstrap/issue/294/docs/2.0/utilities/borders/#border-thickness


danahertzberg:
I'm having issues working with the bootstrap file at /docs/2.0/node_modules/bootstrap/scss/utilities/_borders.scss. It has `!important` tags that need to be removed and I'm not sure how to achieve that.

trackleft:
> I'm having issues working with the bootstrap file at /docs/2.0/node_modules/bootstrap/scss/utilities/_borders.scss. It has `!important` tags that need to be removed and I'm not sure how to achieve that.

copy _borders.scss to here: https://github.com/az-digital/arizona-bootstrap/tree/main/scss
make your override
then add it here: https://github.com/az-digital/arizona-bootstrap/blob/main/scss/arizona-bootstrap.scss#L57-L80

danahertzberg:
> copy _borders.scss to here: https://github.com/az-digital/arizona-bootstrap/tree/main/scss
> make your override
> then add it here: https://github.com/az-digital/arizona-bootstrap/blob/main/scss/arizona-bootstrap.scss#L57-L80

Thanks @trackleft! Did I do everything right? It seems like the original file is still overriding with its darn importance.

scss/_custom-styles.scss
```diff
@@ -246,3 +246,9 @@
     column-count: 4;
   }
 }
+
+// >> Border thick utility.
+
+.border-thick {
+  border-width: .25rem;
```
mmunro-ltrr, line 253: Noticing that the docs say there's an `!important` here, but it is missing. Adding it on a local build produces nice 
thick borders on the example.

danahertzberg: The linting checks didn't like when I had the `!important` here at first. Did your local build remove the `!important` from  `.border`? If that's removed, then this `!important` isn't needed (and other added benefits).

joeparsons, line 253: Something like this should make the stylelint checks happy, unless a better solution exists.
```suggestion
  /* stylelint-disable-next-line declaration-no-important */
  border-width: .25rem !important;
```

mmunro-ltrr: Verified that the local linting checks didn't like adding `!important`: looking next at the order in which the rules are applied (I didn't change anything else).

joeparsons:
FWIW, the original `node_modules/bootstrap/scss/utilities/_borders.scss` is still being imported by the utlilities file included here:
https://github.com/az-digital/arizona-bootstrap/blob/issue/294/scss/arizona-bootstrap.scss#L54

(See https://github.com/twbs/bootstrap/blob/v4-dev/scss/_utilities.scss)

mmunro-ltrr:
And the original Bootstrap `_borders.scss` just escapes the linting errors by turning off `declaration-no-important` for the entire file. Trying to think of an elegant way to override the full utilities import...

danahertzberg:
@joeparsons @mmunro-ltrr I see Arizona Bootstrap doesn't have a Utilities folder that gets imported the same way as TWBS. If we create that same structure, would we be able to do our overrides?

scss/_custom-styles.scss
trackleft: I think if we don't want to import the whole _border.scss file then we should make a file for _custom_overrides like we did for ua-bootstrap 

https://bitbucket.org/uadigital/ua-bootstrap/src/master/src/sass/_custom_overrides.scss

trackleft: Or is this a new class?

mmunro-ltrr:
As a stopgap, could we just put up with duelling `important!`s and not try to override the file in the utilities subdirectory? In the longer term I'm thinking about first documenting everything we're currently overriding or modifying in yet another of those big spreadsheets, then giving some thought to refactoring it. There are a couple of other messy issues that haven'y been formally recorded yet, such as what happens when we've duplicated files from Twitter Bootstrap with some minor changes, then they've made bugfixes in the originals. That should probably be bumped until after General Availability for Quickstart.

trackleft:
> As a stopgap, could we just put up with duelling `important!`s and not try to override the file in the utilities subdirectory? In the longer term I'm thinking about first documenting everything we're currently overriding or modifying in yet another of those big spreadsheets, then giving some thought to refactoring it. There are a couple of other messy issues that haven'y been formally recorded yet, such as what happens when we've duplicated files from Twitter Bootstrap with some minor changes, then they've made bugfixes in the originals. That should probably be bumped until after General Availability for Quickstart.

yeah, i like the idea of adding a custom_overrides file if we aren't overriding everything in the file.

trackleft:
After downloading the pr and testing, it looks like all that is needed is 

_custom_styles.scss
```
// >> Border thick utility.

.border-thick {
  /* stylelint-disable-next-line declaration-no-important */
  border-width: .25rem !important;
}
```

danahertzberg:
Not great to have to battle `!important`s with `!important`s, but this is now ready for review.

scss/_custom-styles.scss
```diff
@@ -89,7 +89,8 @@
 */
 @each $color, $value in $colors {
```
trackleft, line 90: $colors here: https://github.com/az-digital/arizona-bootstrap/blob/main/scss/_custom-variables.scss#L27

trackleft (approved):

scss/_custom-styles.scss
danahertzberg: Right. I don't want to change the colors, just needed to add importance to the border-color style.

trackleft: Totally, just adding for reference

mmunro-ltrr (approved):
Re-examined on a local build.

joeparsons (approved):

PR #314, opened 2021-06-22
Relates to #689

trackleft:
In order make the navigation region span the whole column, we need some addition bootstrap settings.

### DIFF
--- a/scss/_navbar-offcanvas.scss
+++ b/scss/_navbar-offcanvas.scss
@@ -36,8 +36,10 @@
   z-index: 2000;
   display: flex;
   flex-direction: column;
+  flex-basis: 0;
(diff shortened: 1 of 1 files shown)


scss/_navbar-offcanvas.scss
```diff
@@ -36,8 +36,10 @@
   z-index: 2000;
   display: flex;
   flex-direction: column;
+  flex-basis: 0;
```
joeparsons, line 39: stylelint failure:
```
scss/_navbar-offcanvas.scss
 39:3  ✖  Expected "flex-basis" to come before "flex-direction"   order/properties-order
```

```suggestion
  flex-basis: 0;
  flex-direction: column;
```

PR #336, opened 2021-07-22
Closes #334: Update the devDependencies and linting configuration, following Twitter Bootstrap.

### LINKED ISSUE #334: Dev dependencies need routine updates (mid-July, 2021).
Upstream Twitter Bootstrap has bumped the version numbers for several development dependencies, and relaxed some of the eslint rules; Arizona Bootstrap should re-synchronize with these changes, even if they're not yet applied to the upstream v4-dev branch.

## Conditions of satisfaction
- [ ] Should match the dependencies in Twitter Bootstrap and any Dependabot suggestions not ignored there.

mmunro-ltrr:
Upstream updates https://github.com/twbs/bootstrap/commit/5d24fef741944c168ad0ca9cabd1509d5803e441 not only bumped some version numbers, but also revised some of the linting configuration files. The new versions and configuration started failing some of the documentation where it had lagged behind the upstream version (for example, the checks now object to lorem ipsum text in places where a `lang="en"` applies implicitly, but upstream has switched to using semi-meaningless English placeholder text). A very obscure bug disabled local testing: the `node_modules` directory tree included some `.DS_Store` files (a common piece of Mac-specific cruft), despite being generated within a Docker container running on a Linux host. One of these files broke part of the linting process (not by being detected, but by showing up in the rules subdirectory of a plugin module for eslint). A workaround is to clean some of these out in advance.


joeparsons (approved):
Looks OK to me.

trackleft (approved):

Issue #377, opened 2021-10-06
GitHub actions fail to update the CDN correctly.


mmunro-ltrr:
## Problem/Motivation
The CDN in S3 does not reflect changes to the latest built artefacts (on the main branch).

### Describe the bug
The GitHub Actions triggered on a merge to the main branch do update the CDN (modify the AWS S3 bucket, force a Cloudfront cache invalidation), but the files used are stale, not reflecting the newly merged code.

### To Reproduce

1. Merge a PR to main that modifies the generated CSS.
2. Verify the modifications are absent in the CDN at https://cdn.digital.arizona.edu/lib/arizona-bootstrap/main/css/arizona-bootstrap.css

## Proposed resolution
Ensure GitHub Actions make all the expected changes to the CDN on each update to the main branch, and each release.

mmunro-ltrr:
> Question: Do we ever really need to see the diff of the processed CSS?

It would be nice as one of the checks on each commit to a feature branch (even before the merge to main) to catch anything unexpected before it's too late.

mmunro-ltrr:
It's just back to the long-standing problem of what to do with the `/dist` directory subtree. The old way of updating this relied on the modifications getting merged along with pull requests, but Actions were supposed to take over doing this. This works for the review site, but is broken elsewhere. The slightly messy detail will be either going with the old suggestion not to have `/dist` as a permanent directory in the repository at all (.gitignored, & requiring developers to regenerate it in their local environments), or having a GitHub Action rebuild it & automatically commit the result immediately before anything gets merged into the main branch.

PR #379, opened 2021-10-08
Closes #378: Add spacing classes my-20 py-30 to the packaged CSS.

### LINKED ISSUE #378: Spacing classes my-20 py-30 classes are in documentation, but not in packaged css.
## Problem/Motivation
It was discovered that the css required for this was not included.

## Proposed resolution
Run build on main, and commit the css .

mmunro-ltrr:
Bugfix modification, in advance of a more permanent solution to Issue #377.


mmunro-ltrr:
> Should this close both #377 and #378?

The fix for #377 will be a different, much bigger, PR.

PR #492, opened 2022-06-15
Test enabling gradients

danahertzberg:
Testing with automated build


danahertzberg:
Tested. Will not use.

Issue #420, opened 2022-01-19
The code that builds the documentation causes GitHub Action failures with recent dev dependency versions.


mmunro-ltrr:
## Problem/Motivation
Arizona Bootstrap can't follow recommended development dependencies because of a GitHub Action failure.

### Describe the bug
The Hugo static site generator builds the documentation, and a GitHub Action invokes this as part of the linting check that has to pass before we can merge pull requests. Recent versions of the `hugo-bin` dev dependency in package.json cause an error at that point, so currently this is still specifying the 0.78.1 version, while upstream Twitter Bootstrap has already advanced to 0.80.0.

### To Reproduce

Bump the version to 0.80.0, observe a failure in the linting GitHub Action as in
https://github.com/az-digital/arizona-bootstrap/runs/4813153270?check_suite_focus=true
```
Start building sites … 
hugo v0.92.0-B3549403+extended linux/amd64 BuildDate=2022-01-12T08:23:18Z VendorInfo=gohugoio
Error: Error building site: POSTCSS: failed to transform "docs/2.0/assets/css/docs.css" (text/css): SystemError [ERR_SYSTEM_ERROR]: A system error occurred: uv_os_homedir returned ENOENT (no such file or directory)
Total in 617 ms
    at Object.<anonymous> (/usr/local/lib/node_modules/npm/node_modules/clean-stack/index.js:6:61)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/usr/local/lib/node_modules/npm/node_modules/aggregate-error/index.js:3:20)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
ERROR: "docs-build" exited with 255.
Error: Process completed with exit code 1.
```
Annoyingly, the Action is using a Docker container, supposedly to guarantee identical behavior to local dev environments, but they cannot reproduce this failure.

## Proposed resolution

[...]

PR #431, opened 2022-02-25
Closes #420: attempt solutions for Hugo failures when run within a GitHub action.

mmunro-ltrr:
The only way to test the GitHub Action bug is to make this PR.

PR #522, opened 2022-08-02
Closes #521: Bump dev dependency version numbers and refactor Actions.

### LINKED ISSUE #521: Dev dependencies need routine updates (early August, 2022).
## Conditions of satisfaction
- [ ] Dependabot should show fewer proposed updates.
- [ ] The dev dependencies should be similar to upstream twbs/bootstrap v4-dev and main branches.

mmunro-ltrr:
This includes yet another fragile refactoring of the GitHub actions setup, so might need more work before actual review. Some dependabot-suggested version number bumps do not appear, because twbs/bootstrap has ignored them for now.


trackleft:
The changes here are probably ok https://github.com/az-digital/arizona-bootstrap/pull/522/files#diff-1a9dd1a33d61bc61437d2aa16f3d01169a25cb21fa3b8e395d992a1d264cdcca

@akslay @danahertzberg

dist/css/arizona-bootstrap.css
```diff
@@ -6204,7 +6224,7 @@ a.close.disabled {
   width: 2rem;
   height: 2rem;
   vertical-align: -0.125em;
-  border: 0.25em solid currentColor;
+  border: 0.25em solid currentcolor;
```
trackleft, line 6227: Never heard of this but, apparently it is a thing https://css-tricks.com/currentcolor/

mmunro-ltrr:
Unfortunately the review site is almost completely broken (which prompted https://github.com/az-digital/arizona-bootstrap/issues/523). I didn't want to relegate this PR to a WIP, though (wanting to test any updates as changes to an active PR). The breakage has no connection with currentcolor, but the baseURL not getting included in many of the generated links.

mmunro-ltrr:
This should really be ready to merge, after the latest changes. Invalid relative URLs caused the broken review site; they had been present all along, but a bug in Hugo was letting them work: when this was fixed in the v101 release, the existing problem started manifesting itself. The most recent commit with version number bumps actually changes the included version of upstream twbs Bootstrap to 4.6.2, with the main consequence being extra noise in the diffs (from updated version number comments).

A big hope of this PR is that it will finally close https://github.com/az-digital/arizona-bootstrap/issues/420 and obsolete the draft https://github.com/az-digital/arizona-bootstrap/pull/431 allowing Arizona Bootstrap to use current versions of Hugo to build the documentation files that are also used in some of the linting checks. Although GitHub Actions were using a Docker image built from the same Dockerfile used in local dev environments, the default user in the container wasn't the `node` user defined in the image, but one that was a default in the GitHub Action runner. A routine update to Hugo changed one of its internal dependencies to a version that tried to access the home directory of the user under which it was running, but no such home directory existed within the container. Forcing more operations within the container to run as the `node` user, and fixing the file ownership to prevent this conflicting with the default GitHub Action user seems to be working, but a follow-up check will be that local dev environments are not affected (my local setup has the default user outside the container using the same uid as the `node` user inside, so is a bad test case).

joeparsons (approved):

dist/css/arizona-bootstrap.css
bberndt-uaz: Yes, it looks like `currentcolor` is correct: https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword

trackleft (approved):

PR #303, opened 2021-06-09
Closes #291 Updating accordion styles

smayden:
This PR makes two changes to the accordion styles in Bootstrap:
- Adds text-transform: none to the _collapse.scss file to keep the accordion headings regular casing not uppercase
- Adds the arrows on the accordion with the Material Icons Sharp

In testing on the review site, I did not see the reported overlap issue between long headings and the arrows. That may need to be addressed separately in Quickstart.

image

Review site: https://review.digital.arizona.edu/arizona-bootstrap/issue/291/docs/2.0/components/collapse/

Thanks to @danahertzberg for help during the workshop on how to do bootstrap changes and design input.

PR #532, opened 2022-08-12
Closes #537: Restyle breadcrumbs

### LINKED ISSUE #537: Improve breadcrumb display
## Motivation
I want to distinguish breadcrumb navigation from page content.

## Solution
- [ ] ~Make text size small~
- [ ] Reduce breadcrumb padding and margin to settings below
     - [ ] Reassess padding after other changes
- [ ] ~Add gray-200 background color behind links (not full width)~
- [ ] Remove any hard-coded colors if not black (allow bg-gray-200 to set color)
- [ ] Change slashes to chevron right GM icon (content: “\e5cc”;)
- [ ] Add home icon (content: “\e88a”;)
- [ ] Update bootstrap documentation with recommendations on how to present breadcrumbs on your site.

```
.breadcrumb-item+.breadcrumb-item::before {
    float: left;
    padding-right: 0.5rem;
    color: #000000;
    content: “\e5cc”;
    font-family: ‘Material Icons Sharp’;
}
.breadcrumb {
    font-size: small;
    display: inline-flex;
    padding: 2px 5px;
    margin: 5px 0;
}
.breadcrumb-item.active {
-/* color: #6c757d; */-
}

Home icon
     content: “\e88a”;
     font-family: ‘Material Icons Sharp’;
```

## Context
https://docs.google.com/document/d/1FnDTe3tBeK31wKyTdlujoel9d6X9XBT63s3vO_U0zEw/edit#heading=h.zhkufj9wt9vw

cpoff:
**Updated**: We need agreement on what the base breadcrumb contains. Then we can seek agreement on the design and implementation of this ticket.

ORIGINAL: This adds background color (gray-200), a home icon, and changes the divider to a right chevron. It also uses the fit-content attribute to prevent the breadcrumb from spanning full width. We also re-ordered the import sequence.

Chris walked me through most of the steps, including adding custom variable overrides. Sorry if it's a bit klunky right now but I'm happy to clean it up as I get more familiar.


scss/arizona-bootstrap.scss
```diff
@@ -10,6 +7,8 @@
  */
 @import "../node_modules/bootstrap/scss/functions";
 @import "../node_modules/bootstrap/scss/variables";
+@import "custom-variables";
```
trackleft, line 10:
```suggestion
```

scss/arizona-bootstrap.scss
```diff
@@ -1,7 +1,4 @@
-@import "custom-variables";
-
 // Required Bootstrap imports.
```
trackleft, line 1:
```suggestion
@import "custom-variables";

// Required Bootstrap imports.
```

trackleft (changes requested):
I think that custom-variables needs to be where it was.

mmunro-ltrr:
Once the whitespace issues are cleaned up, the linting starts complaining about unused variables (seeing this in local testing).

mmunro-ltrr:
The changes I pushed directly to the https://github.com/az-digital/arizona-bootstrap/tree/issue/1825-breadcrumbs branch are just to get the review site working again, but more needs to be done. Some suggestions:
 - The `scss/_variable-overrides.scss` file isn't used by anything apart from this breadcrumb override, so we can delete it.
 - The actual breadcrumb component overrides are in `scss/_pagination.scss`, so the references to the modified variables can go there.
 - The variable value overrides removed from `scss/_custom-variables.scss` to shut up the linting check for unused variables can go back, because they're now used somewhere.

scss/arizona-bootstrap.scss
```diff
@@ -9,6 +9,7 @@
  */
 @import "../node_modules/bootstrap/scss/functions";
 @import "../node_modules/bootstrap/scss/variables";
+@import "variable-overrides";
```
mmunro-ltrr, line 12: The previous suggestion was:
The `scss/_variable-overrides.scss` file isn't used by anything apart from this breadcrumb override, so we can delete it.
So it shouldn't be referenced, and the file itself deleted.

scss/_custom-variables.scss
```diff
@@ -228,6 +228,11 @@ $spacers: map-merge(
 $badge-pill-radius: .25rem;
 
 // >> Breadcrumbs
+$breadcrumb-font-size: small;
```
mmunro-ltrr, line 231: The previous suggestion:
“The variable value overrides removed from `scss/_custom-variables.scss` to shut up the linting check for unused variables can go back, because they're now used somewhere.”
is contingent on them _actually_ being used, so they can't be added without modifying `scss/_pagination.scss`

mmunro-ltrr (changes requested):

scss/_pagination.scss
```diff
@@ -5,7 +5,20 @@
...
   background-color: $breadcrumb-bg;
+
+
+  &::before {
+    float: left; // Suppress inline spacings and underlining of the separator
+    padding-right: $breadcrumb-item-padding;
+    // stylelint-disable-next-line
+    font-family: "Material Icons Sharp";
+    color: $breadcrumb-divider-color;
+    content: "\e5cc";
```
mmunro-ltrr, line 20: We should follow the pattern of
https://github.com/az-digital/arizona-bootstrap/pull/303
(with a name rather than a codepoint, for example).

scss/_pagination.scss
```diff
@@ -6,9 +6,21 @@
 //
 .breadcrumb {
   background-color: $breadcrumb-bg;
+
```
trackleft, line 9: Should this be optional?  Don't ask me.

scss/_custom-styles.scss
```diff
@@ -5,6 +5,15 @@
 /*
 * >> Background Color
 */
```
trackleft, line 7: Question, does overriding twbs classes go in _custom-styles.scss  in `_breadcrumb.scss` like it is in https://github.com/twbs/bootstrap/blob/v4.6.0/scss/_breadcrumb.scss

scss/_pagination.scss
trackleft: Same question about this class being in pagination.scss vs breadcrumb.scss

trackleft: Sorry if these are annoying questions ha

scss/_pagination.scss
```diff
@@ -6,9 +6,21 @@
...
+  &::before {
+    float: left; // Suppress inline spacings and underlining of the separator
+    padding-right: $breadcrumb-item-padding;
+    // stylelint-disable-next-line
+    font-family: "Material Icons Sharp";
```
trackleft, line 14: Now that I see the font-family, I think these icons should be optional, since material icon font is not explicitly required in arizona bootstrap.

scss/_pagination.scss
cpoff: `breadcrumb.scss` is under Node Modules. And according to Martin: "Nothing that's under node_modules should get edited: that's the directory tree where npm puts all the dependencies, including the unmodified upstream Bootstrap. So they don't show up in the https://github.com/az-digital/arizona-bootstrap view of the repository at all (the .gitignore file explicitly block changes to node_modules)."

scss/_custom-styles.scss
trackleft: Like this I mean https://github.com/az-digital/arizona-bootstrap/pull/538/files

mmunro-ltrr:
Part of the confusion is the messy combination of overrides for both upstream Twitter bootstrap `_breadcrumb.scss` and `_pagination.scss` in just Arizona Bootstrap's `_pagination.scss`... and as I type this I see https://github.com/az-digital/arizona-bootstrap/pull/538 from @trackleft to address the issue.

trackleft:
I am not 100% sold, that my PR is a better approach, but my suspicion is that separating our overrides into files that mirror the source twbs file structure could make it easier down the road for us to build components in a design system with required bootstrap css kept at a minimum.

At the end of the day we would like to be able to only add the required JS/CSS to any given page, and leave out all other JS/CSS. 

The way we do it now, in Quickstart, is to include the whole kit & kaboodle, but that has the possibility of changing with CKEditor 5 and the limiting straight HTML fields even more. <-- Probably an impossible dream, but still want to try.

All of this effort, to improve performance across all sites one day.

cpoff:
This should now be ready for review. 

http://localhost:9001/docs/2.0/components/breadcrumb/

mmunro-ltrr (dismissed):
Looks OK, if this is the desired design! A couple of weirdnesses: a section header comment (for Background Color) has gone from `_custom-styles.scss` & the supposed changes in `font.md` are just what's in the current main branch (so those don't matter).

danahertzberg (approved):

mmunro-ltrr (approved):

camikazegreen:
Bringing some conversations from Slack into this ticket:

[Dana Hertzberg]
Hi Marcom team! Do you have any concerns or insights on these breadcrumb icon changes?
Current: https://digital.arizona.edu/arizona-bootstrap/docs/2.0/components/breadcrumb/
New: https://review.digital.arizona.edu/arizona-bootstrap/issue/1825-breadcrumbs/docs/2.0/components/breadcrumb/

[Thomas McCarthy]
Hi Dana! I personally like the switch to the chevrons. Feels more action orientated  - like moving forward on a path. I was curious what the motivation was to add the home icon and still keep the "Home" title? Is that for screen readers? If we are to keep both, I would make the home icon Azurite as well.

[Catherine Gustafson]
No concerns, I also like the switch to chevrons. If there is a design change, our team can update instances of breadcrumbs being used where the first link is not the homepage link. The Purpose, Mission & Values pages for instance would need to be updated to include a ‘home’ link (this may be best practice anyways :slightly_smiling_face: , so a good reminder)

[Dana Hertzberg]
Thanks 
The home icon further helps identify the content as a breadcrumb trail with a recognizable icon.

[Thomas McCarthy]
Thanks 

 I was just wondering if having both (icon + "home") is necessary. I thought the motivation of using the icon might be to reduce the size of the breadcrumbs. The home icon is considered a universal icon, maybe that's all we need?

[Dana Hertzberg]
Hi! The motivation behind icon and text is to have a larger tap/click area for the link while also enhancing the recognizability of breadcrumbs.

[Thomas McCarthy]
Okay, I personally feel the icon is enough but I don't have any research to back that up

trackleft:
This is still waiting on documentation updates.

cpoff:
We'd like to consider some baseline best practices for breadcrumb discussion in collaboration with the broader group. Just to make sure our implementation hits the desired/agreed upon targets.

Examples:
- The default breadcrumb will/will not contain a home icon
- Will/will not contain the word HOME
- Will/will not have active page greyed out
-

camikazegreen:
Noting my preferences here to Curt's questions:

-     The default breadcrumb **will** contain a home icon
  -     Will it be azurite or grey? My preference is azurite, but is dependent on whether or not it's clickable.
  -     Will it be clickable? My preference is yes
-     **Will not** contain the word HOME
-     **Will** have active page greyed out

mmunro-ltrr:
The merge conflicts are because https://github.com/az-digital/arizona-bootstrap/pull/535 changed some files on the `main` branch. Rather than fix them manually, first merging the current state of `main` into the `issue/1825-breadcrumbs` branch used by this PR should resolve the conflicts.

danahertzberg:
If home icon always exists, there should be a chevron in front of any non-home breadcrumb. There should not be a chevron between the home icon and front page breadcrumb.

Could put the home icon in a new class and apply that class only to the front page breadcrumb. Then the home icon will only exist if the front page breadcrumb exists based on site configuration.

mmunro-ltrr:
Unfortunately the merge conflicts are still there.

cpoff:
Thanks, Martin. Ready for re-review.

mmunro-ltrr:
Just re-reading previous comments & looking at a fresh build: if the requirement is to have the home icon clickable on the first link, and omit any text like “Home”, that will need some further refactoring. Was there a consensus on this, or is the current appearance OK for now?

danahertzberg:
> Just re-reading previous comments & looking at a fresh build: if the requirement is to have the home icon clickable on the first link, and omit any text like “Home”, that will need some further refactoring. Was there a consensus on this, or is the current appearance OK for now?

@mmunro-ltrr @cpoff The current appearance is good for now. But I recall discussing having the home icon appear only if the first breadcrumb item links to the front page/is "Home".

Let's add the home icon CSS to a new class that's applied just to the home breadcrumb item. Then we'll need a follow up ticket in QS to add that class if the home breadcrumb is displayed.

mmunro-ltrr (dismissed):
Good to go for now, but a follow-up is definitely needed for the comment from @danahertzberg
The documentation at `site/content/docs/2.0/components/breadcrumb.md` needs a little tweak in addition to any changes demonstrating & documenting the breadcrumbs themselves: the link to the ARIA guidelines specifically for breadcrumbs is outdated, and now
https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/
(but that can go into the follow-ups).

danahertzberg:
- [ ] Move front page before CSS to new `breadcrumb-home` class
- [ ] (Quickstart) Add new `breadcrumb-home` class to front page segment if "Include the front page as a segment in the breadcrumb" option is checked

scss/_breadcrumb.scss
```diff
@@ -0,0 +1,38 @@
...
+    padding: 1px 0;
+    color: $breadcrumb-color;
+
+    &:hover {
+      text-decoration: none;
+      border-bottom: 2px solid $azurite;
+    }
+  }
+}
+
```
danahertzberg, line 30:
```suggestion
  a {
    padding: 1px 0;
    color: $breadcrumb-color;
    &:hover {
      text-decoration: none;
      border-bottom: 2px solid $azurite;
    }
  }
}
.breadcrumb-item-home::before {
  float: left; // Suppress inline spacings and underlining of the separator
  padding-right: $breadcrumb-item-padding;
  // stylelint-disable-next-line
  font-family: "Material Icons Sharp";
  color: $breadcrumb-divider-color;
  content: "home";
}

```

scss/_pagination.scss
```diff
@@ -1,26 +1,4 @@
 // Overrides for Pagination, Breadcrumbs
```
danahertzberg, line 1:
```suggestion
// Overrides for Pagination
```

mmunro-ltrr (dismissed):

danahertzberg (approved):

trackleft (approved):

site/layouts/shortcodes/alert.html
```diff
@@ -0,0 +1,16 @@
+{{- /*
+  Usage: `placeholder args`
```
joeparsons, line 2:
```suggestion
  Usage: `alert args`
```

danahertzberg (approved):

trackleft (approved):

joeparsons (approved):

mmunro-ltrr (approved):

Issue #536, opened 2022-08-19
Review sites are not deployed from some pull requests


mmunro-ltrr:
## Problem/Motivation
Review the history of GitHub Actions on this repository: many deployments to the AWS review site fail at an early stage, when handling the `package-lock.json` file.

### Describe the bug
In many cases the lockfile should be unmodified, but committing a changed version causes a permission error.

## Proposed resolution
Investigate the use of the `npm ci` command in place of `npm install` in some contexts.
https://docs.npmjs.com/cli/v8/commands/npm-ci

joeparsons:
Closing this for now since we haven't seen this problem lately.

PR #565, opened 2022-09-21
Closes #563: Copy the file ownership fix-up from the Action that builds review sites on PRs.

mmunro-ltrr:
The Action depends on https://github.com/az-digital/digital.arizona.edu/pull/36

PR #566, opened 2022-09-22
Remove unnecessary(?) git restore package-lock.json step.

joeparsons:
This PR is targeting @mmunro-ltrr's PR branch (#565). 

@mmunro-ltrr I might be missing something obvious but it seems to me that the `git restore package-lock.json` step that is failing in the review site actions workflow on your new PR might not be necessary at all (?).

### DIFF
--- a/.github/workflows/review-site.yml
+++ b/.github/workflows/review-site.yml
@@ -123,9 +123,6 @@ jobs:
         run: |
           git config --global user.email "${GITHUB_ACTOR}@users.noreply.github.com"
           git config --global user.name "${GITHUB_ACTOR}"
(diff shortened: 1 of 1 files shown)


mmunro-ltrr:
The intent here is to prevent any update of the `package-lock.json` file that's included in pending changes to be pushed back from the cloned copy that the GitHub Action has been using to the main repository. But if it's not staged with the rest of the files that should already prevent an unwanted update. This might also fix https://github.com/az-digital/arizona-bootstrap/issues/536

mmunro-ltrr (approved):
This is worth a try. One other thought had been dropping support for a fully committed lockfile altogether, but that's ugly.

PR #611, opened 2022-11-23
Closes #610 Update navbar to account for future versions.

### LINKED ISSUE #610: Update navigation on Arizona Bootstrap docs to account for version 5
Update navbar version links.

trackleft:
The goal should be to provide a way to find the latest version, and to know which version you are on currently.
Additionally, we don't want to update all documentation pages for all versions upon each release, so our wording and links need to be sufficiently generic or reuseable.


trackleft:
For final patch release prior to new major/minor release remove version number from **Latest (2.0.19)**
Set latest url to permanent [latest path](https://digital.arizona.edu/arizona-bootstrap/)

Each new release will add to the menu
Previous releases are frozen in time

site/layouts/partials/docs-navbar.html
```diff
@@ -19,12 +19,13 @@
         v{{ .Site.Params.docs_version }}
       </a>
       <div class="dropdown-menu dropdown-menu-md-right" aria-labelledby="bd-versions">
-        <a class="dropdown-item active" href="{{ `/` | relURL}}">Latest (2.0.x)</a>
+        <a class="dropdown-item text-white" href="{{ `/` | relURL }}">Arizona Bootstrap 5.0.x</a>
```
trackleft, line 22:
```suggestion
        <a class="dropdown-item text-white" href="{{ `/` | relURL }}">Latest</a>
```

mmunro-ltrr (approved):
There might be more changes as we re-do the docs (changes to the look & feel of the 2.x docs as well as the changes) for 5.x, but it's good to get this in first. Tested locally, too.

joeparsons (approved):

tadean (approved):

PR #773, opened 2023-05-17
Closes #736: Include April and May 2023 Dependabot dev dependency updates.

### LINKED ISSUE #736: Early April Dependabot Arizona Bootstrap development dependency pull requests need cleaned up.
Dependabot is generating many pull requests for the development dependencies of Arizona Bootstrap, making it difficult to prioritize important issues. An intended change was to allow direct merging of these pull requests, with few if any code changes required (just updating some of the GitHub repository secrets), but if this cannot be done soon, a manual combined dependency update is needed as a housekeeping measure. Ideally this would be before the next patch release of Arizona Bootstrap,which in turn would be in advance of the next release in az_quickstart.

mmunro-ltrr:


mmunro-ltrr:
The issue/736 branch is missing a commit on main: regenerating it, and making a new PR.

PR #899, opened 2023-08-22
Closes #887: Address off-canvas menu usability issues on iOS

### LINKED ISSUE #887: Navigation not clickable on Mobile
## Problem/Motivation
Navigation on mobile cannot be reached when the subnavigation has too many items. 

### Describe the bug
I came across this on a development site, I tried to scroll down to the last item in the Mobile navigation and had a hard time getting navigation to show on the page once its does show it cannot be clicked.

### To Reproduce
video example: https://app.screencast.com/pDTc8mB1X2JSD

1. **On an actual mobile device** ( i used iphone 13pro /safari ) goto: https://dev-marcom-redesign.pantheonsite.io/brand-guidelines/videography/filming-campus
2. Click on 'menu' icon in upper right hand corner
3. Click on 'Our Team' at the bottom of the list
5. try to scroll down to "Cross-Campus Leadership"  if you are able to see it try to click on it - https://app.screencast.com/CdwrxSqvunHU7

## Proposed resolution
Description of the proposed solution, the rationale behind it, and workarounds for people who cannot use the patch.

### Expected behavior
Easy to navigate to last items. see Arizona.edu for expected behavior

##  Screenshots
video example: https://app.screencast.com/pDTc8mB1X2JSD

### Smartphone (please complete the following information):
 - Device: [iPhone 13 pro]
 - OS: [iOS 16.6]
 - Browser [chrome and safari]
 - Version [chrome 115 / Safari 17]

joeparsons:
Closes #887

- Updates `.offcanvas-toggle-body-freeze` class (applied to body when offcanvas nav is activated) to use `overflow: hidden` instead of `overflow-y: hidden` to minimize background scroll issues affecting iOS browsers.
	- I also experimented with adding `position: fixed` to the class an alternative fix but decided it would be better to not reset the scroll on the page any time an off-canvas nav is activated even though this does seem to eliminate background / body scrolling on iOS
	- I could be convinced to add `position: fixed` back if that seems acceptable to everyone
- Changes the`.navbar-offcanvas` container `height` and `max-height` styles  to use the new/draft [small viewport unit](https://web.dev/viewport-units/) (`svh`) to fix the usability of links that requiring scrolling to the bottom of the nav to access
  - I also experimented with using  dynamic viewport units (`dvh`) which responds better to changing browser address bar sizes, etc, but it using small viewport units seems to be safer from an accessibility/performance perspective (according to MDN and other resources)
- I also added some additional links to the example off-canvas navbar in the Bootstrap docs (per @HenryzGreenberg's suggestion) to make this easier to test

This definitely needs testing on as many real mobile devices as possible.

### Before:
https://digital.arizona.edu/arizona-bootstrap/docs/2.0/components/navbar-off-canvas/

### After:
https://review.digital.arizona.edu/arizona-bootstrap/issue/887-bg-scroll/docs/2.0/components/navbar-off-canvas/

## References
- https://webkit.org/blog/12445/new-webkit-features-in-safari-15-4/
- https://web.dev/viewport-units/
- https://caniuse.com/viewport-unit-variants
- https://bugs.webkit.org/show_bug.cgi?id=153852
- https://stackoverflow.com/questions/3047337/does-overflowhidden-applied-to-body-work-on-iphone-safari


scss/_navbar-offcanvas.scss
```diff
@@ -3,6 +3,7 @@
   display: none;
 }
 .offcanvas-toggle-body-freeze {
+  position: fixed;
   overflow-y: hidden;
```
joeparsons, line 7: Looks like upstream Bootstrap 4.x just sets `overflow: hidden` and does not set position fixed which also seems to work with Safari 🤷🏻‍♂️.
```suggestion
  overflow: hidden;
```

mmunro-ltrr (dismissed):
I fired up our az-digital@web.arizona.edu browserstack.com account (there's only one account with live testing enabled). The before & after versions performed as expected, Also verified that a local build passed the basic linting checks. BrowserStack seems to show different behavior with different iPhones: this was with iPhone 13pro /safari .

site/content/docs/2.0/components/navbar-off-canvas.md
```diff
@@ -51,24 +51,44 @@ Adding the `.navbar-offcanvas` class to your `<nav>` will incorporate an off-can
...
-        <a class="dropdown-item" href="#">Something else here</a>
+        <a class="dropdown-item" href="#">Item 5</a>
+        <a class="dropdown-item" href="#">Item 6</a>
+      </div>
+    </li>
+    <li class="nav-item nav-item-parent dropdown keep-open">
+      <button class="nav-link dropdown-toggle" id="navbarDropdown5" data-toggle="dropdown" data-display="static" aria-expanded="false">
+        Third top-level dropdown
+      </button>
+      <div class="dropdown-menu" role="menu" aria-labelledby="navbarDropdown5">
```
joeparsons, line 84:
```suggestion
      <div class="dropdown-menu" role="menu" aria-labelledby="navbarDropdown6">
```

HenryzGreenberg:
Reviewed test URL on Browserstack in multiple devices and browsers. I also checked on my personal devices. This seems to now render as expected.

joeparsons:
Aside:  I temporarily added @danahertzberg's padding-bottom changes to this PR branch but ended up reverting those changes because when I test that in combination with my changes the content of the offcanvas-nav scrolled "behind" the padding at the bottom and it didn't really help the overall issue:
IMG_7240

HenryzGreenberg (approved):
Reviewed test URL on Browserstack in multiple devices and browsers. I also checked on my personal devices. This seems to now render as expected.

mmunro-ltrr (approved):

scss/_navbar-offcanvas.scss
```diff
@@ -38,8 +38,8 @@
   flex-direction: column;
   width: 325px;
   max-width: 90%;
-  height: 100vh;
-  max-height: 100vh;
+  height: 100svh;
+  max-height: 100svh;
```
joeparsons, line 42:
```suggestion
  height: 100vh;
  height: 100svh;
  max-height: 100vh;
  max-height: 100svh;
```

danahertzberg (approved):

mmunro-ltrr (approved):

trackleft:
Let's test the impact of this on core web vitals. I wonder if this is a real concern. 

I do know that the mobile nav renders  on initial page load and then moves off screen, which is probably a bigger problem.

If that were not the case, then I would say that we could set the height dynamically.

mmunro-ltrr (approved):

trackleft (approved):

PR #956, opened 2023-09-28
Create ckeditor-arizona-bootstrap.scss artifact

trackleft:


mmunro-ltrr:
Just tried re-triggering the Action with debug logging enabled. The PR isn't the problem, the Action configuration is getting problematic.

PR #1072, opened 2023-12-12
Update links to UA Brand site

zsmeyer:
Hello!

You may already be aware, but just in case: 
I was looking that the color information for Arizona bootstrap, and when I clicked the link to go to the UA Brand page, I got a cert error.

It looks like maybe they renamed from brand.arizona.edu to marcom.arizona.edu? I think this should update references on the site from the old domain to the new one.

I'll reach out to them next, and I can update this PR with their response if you'd like.

Thanks!
Zach

### DIFF
--- a/site/content/docs/5.0/examples/navbar/index.html
+++ b/site/content/docs/5.0/examples/navbar/index.html
@@ -36,7 +36,7 @@ title: Navbar template
                     <a href="https://quickstart.arizona.edu/" class="text-muted nav-link nav-link-https--quickstartarizonaedu-">Utility 1</a>
                   </li>
                   <li class="nav-item">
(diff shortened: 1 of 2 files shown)


zsmeyer:
Opened https://github.com/az-digital/arizona-bootstrap/issues/1073 for this.

mmunro-ltrr:
This won't work in its present form, because the pull request uses a fork of the arizona-bootstrap repository rather than a branch within the actual repository, but I'll re-roll it for you as a couple of usable PRs.

trackleft:
Addressed in 
https://github.com/az-digital/arizona-bootstrap/pull/1075
https://github.com/az-digital/arizona-bootstrap/pull/1076

PR #1378, opened 2025-02-26
Remove collapse overrides

trackleft:


djcelaya:
Hi, @trackleft. I've been working on the accordion/collapse components recently and took a quick peek at this. I'm not sure it's necessary anymore with the new docs. Do you agree? I have already removed our old collapse overrides as part of the accordion upgrade. Should we close this or leave it open for further consideration?

PR #1430, opened 2025-04-23
Closes #1429: Generate colors using the Utility API

bberndt-uaz:
WIP

Switch our overrides to use the Bootstrap 5 utility API for generating colors, where possible.

Also adds `text-bg-` classes (to replace our `bg-` classes in Quickstart).

Issue #1419, opened 2025-04-18
Update border colors to reflect brand for Arizona Bootstrap 5


danahertzberg:
Update border colors to reflect brand https://getbootstrap.com/docs/5.0/utilities/borders/#border-color

- Primary, Neutral, and Complimentary
- https://marcom.arizona.edu/brand-guidelines/colors

Issue #1417, opened 2025-04-18
Update background colors for Arizona Bootstrap 5


danahertzberg:
Update background colors https://getbootstrap.com/docs/5.0/utilities/background/

Match existing https://digital.arizona.edu/arizona-bootstrap/docs/2.0/utilities/colors/#background-color

PR #1443, opened 2025-04-28
Closes #1401, #1417, and #1419: Add text, border, and background colors with opacity using Utilities API. Add brand colors as RGB CSS custom properties

### LINKED ISSUE #1401: Update text colors for Arizona Bootstrap 5
## Conditions of satisfaction

- Colors provided by upstream Bootstrap: https://getbootstrap.com/docs/5.3/customize/color/
- Arizona Bootstrap 5 colors should match 2.0: https://digital.arizona.edu/arizona-bootstrap/docs/2.0/utilities/colors/

kevdevlu:
Includes Brian's #1430 

Adds Arizona brand colors to texts, backgrounds, and borders using utilities API 

1. Adjusted the existing code to align with this new bootstrap functionality: 

    a. standard background `.bg-{color}` classes do not adjust text `color` to contrast with background color

    b. the new helper `.text-bg` classes _do_ adjust text `color` for optimal text-background contrast (https://getbootstrap.com/docs/5.3/helpers/color-background/)

3. Text, Background, and Border colors:

    a. Arizona branded colors are converted into RGB and added as CSS custom properties. 

    b. These CSS custom properties are used for text, background, and border colors and are added into the `$utilities` variable along with their corresponding opacity variables (`--bs-text-opacity` for text, `--bs-bg-opacity` for background, and `--bs-border-opacity` for border)

    c. this means we can add `.bg-opacity-*` , `.text-opacity-*`, and `.border-opacity-*` to their corresponding class and expect opacity to obey where * is a number in {10, 25, 50, 75, 100} 

5. added opacity capability (**2c**) to new helper `.text-bg` classes (**1b**) 

### Follow-up issue(s)
 - The MarCom Colors page says that [highlight/accent colors should not be used as text colors](https://marcom.arizona.edu/brand-guidelines/colors). Potentially deprecate or remove bloom, sky, leaf, river, and mesa as text colors
 - Investigate changing the prefix for our custom color properties from `bs` to `az` (e.g., `--az-red-rgb`)
 - Investigate using upstream Bootstrap white, black, and grays instead of declaring our own versions of these variables


bberndt-uaz:
@kevdevlu Thanks for putting this all together!

I think the one thing I would recommend for this pull request is adding the brand colors to the Color and Background page at `/docs/5.0/helpers/color-background/`.

Otherwise, I wanted to note a few things (which aren't blockers):
 - (For @danahertzberg in particular) I noticed that the MarCom Colors page says that [highlight/accent colors should not be used as text colors](https://marcom.arizona.edu/brand-guidelines/colors). Should we deprecate or remove bloom, sky, leaf, river, and mesa as text colors? 
 - I created a separate issue for restoring the Color Contrast page: #1464.
 - I wish we could use the upstream variables for white, black, and the grays instead of declaring them as custom variables, but it seems necessary since we use them in other custom variables right now. Maybe in the future we can find a way to remove our versions of those variables.

mmunro-ltrr (approved):

joeparsons (approved):

PR #1435, opened 2025-04-24
Closes #1434: Merge the v5-docs branch into main

bberndt-uaz:
Updates the documentation site to be based on upstream Bootstrap 5 docs.

Numerous follow-ups to come to apply/re-apply branding and other changes.

Remaining issues:

- [x] Resolve numerous warnings with the format: `No name was provided for external module "/arizona-bootstrap-source/node_modules/bootstrap/js/src/base-component.js" in "output.globals" – guessing "BaseComponent".`
- [x] Resolve rollup warnings for unresolved dependencies (`bootstrap/js/src/base-component.js (imported by "js/src/offcanvasmenu.js")`)
- [x] Resolve rollup warning for missing global variable name (`Use "output.globals" to specify browser global variable names corresponding to external modules: bootstrap/js/src/base-component.js (guessing "BaseComponent$1")`)

Questions:

1. Should we revert the V5_AUDIT changes to `hugo_template.yml` (or rename those values)?
   - (Apr 29, 2025) I tried removing the V5_AUDIT text and keeping the new params in the Hugo template. However, I reverted these changes since removing the line that added those files to the vnu-jar ignores resulted in many errors like this:
> "file:/arizona-bootstrap-source/_site/docs/5.0/utilities/shadows/index.html":624.5-624.101: error: Bad value "js_bundle_hash" for attribute "integrity" on element "script": Values must start with "sha256-" or "sha384-" or "sha512-".

PR #1448, opened 2025-04-29
Add rollup commonjs plugin, delete upstream js/src and js/tests files, fix Javascript docs page.

joeparsons:
Seems to fix the `No names provided for external module` warnings happening in @bberndt-uaz's [PR branch](https://github.com/az-digital/arizona-bootstrap/pull/1435):
>Resolve numerous warnings with the format: No name was provided for external module "/arizona-bootstrap-source/node_modules/bootstrap/js/src/base-component.js" in "output.globals" – guessing "BaseComponent".

I also removed the upstream Bootstrap files from `js/src` and `js/tests` because I'm hoping we won't have to maintain duplicate copies of those.  The only negative side effect of removing those that I saw was a Hugo error affecting a single docs page which I was able to fix but changing a file path reference in that documentation file (the Javascript page):
```diff
-{{< js-docs name="allow-list" file="sanitizer.js" jsroot="js/src/util" >}}
+{{< js-docs name="allow-list" file="sanitizer.js" jsroot="node_modules/bootstrap/js/src/util" >}}
```


bberndt-uaz (approved):
Thank you @joeparsons! I'll go ahead and merge this into my branch.

Issue #1418, opened 2025-04-18
Update heading colors to AZ Blue for Arizona Bootstrap 5


danahertzberg:
Update heading colors to AZ Blue https://getbootstrap.com/docs/5.0/content/typography/

- Future change to Garamond coming

PR #1451, opened 2025-04-30
Closes #1402, #1418: Update font and typography for Arizona Bootstrap 5

### LINKED ISSUE #1402: Update font and typography for Arizona Bootstrap 5
Typography in upstream Bootstrap: https://getbootstrap.com/docs/5.3/content/typography/

## Conditions of satisfaction
- [ ] Update font to match Arizona Bootstrap 2: https://digital.arizona.edu/arizona-bootstrap/docs/2.0/content/font/ 
- [ ] Proxima Nova Condensed option when marketing guidelines are available?
- [ ] ~Change body text color to AZ Blue~ Keep standard Bootstrap body color
- [ ] Update bold and italic styles to use specified fonts
     - [ ] None: Proxima Nova Regular
     - [ ] **B**: Proxima Nova Bold
     - [ ] _I_: Proxima Nova Italic
     - [ ] **B** & _I_: Proxima Nova Bold Italic

bberndt-uaz:
**This PR makes the following changes:**

1. Adds a "reboot" scss file in the overrides directory and moves `<a>` font-weight styling from _typography.scss to that file
2. Removes our `<body>` color override
3. Cleans up heading styles (removes `.heading-style` and `$font-family-headings`, sets `$headings-font-family` default variable)
4. Changes heading color to Arizona Blue
5. Adjusts blockquote styling to account for `.blockquote-footer`
6. Removes styling for `.blockquote-reverse` (this was dropped in Bootstrap 4)
7. Deletes custom typography style for `.mailto` (this was deprecated in Arizona Bootstrap 2) 
8. Updates the Font docs page and adds it to the sidebar
9. Updates the Typography docs page
10. Removes some unnecessary custom docs styling for headings, adds styling to restore normal font weight for sidebar menu links
11. Deletes the custom `.sans` class
12. Updates the Backwards Compatibility page to note the classes removed

**Questions:**

1. Should we keep, deprecate, or remove `.heading-style` [from our custom styles](https://github.com/az-digital/arizona-bootstrap/blob/main/scss/custom/_styles.scss#L108-L112)?
   - This class is not used in the Quickstart codebase. If we keep it, we should update it and rename it to `.az-heading-style`.
   - (May 6, 2025) [See comment below](#issuecomment-2854717751): we will remove this class.
2. Should we keep, deprecate, or remove `.sans` [from our custom styles](https://github.com/az-digital/arizona-bootstrap/blob/main/scss/custom/_styles.scss#L108-L112)?
   - This class assigns the same font-family we use for body and heading text. We use this class in one place in Quickstart, in the demo content (it can be easily removed). If we keep the class in Arizona Bootstrap 5, we should remove or separate out the font-weight setting.
   - Quickstart includes a CKEditor "Sanserif Text" style which uses an incorrect class name, so it has no effect. I'll create a Quickstart issue to remove or update this style.
   - (May 6, 2025) [See comment below](#issuecomment-2854717751): we will remove this class.
3. Should we keep or deprecate `.text-hyphen` [from our custom styles](https://github.com/az-digital/arizona-bootstrap/blob/main/scss/custom/_styles.scss#L120-L123)?
   - See https://digital.arizona.edu/arizona-bootstrap/docs/2.0/utilities/text/#hyphenate-word in the 2.0 docs and the [PR which added this class](https://github.com/az-digital/arizona-bootstrap/pull/310). Since this class was included in the docs, I think we should not remove it without deprecating it first.
   - This class is not used in Quickstart.
   - (May 6, 2025) [See comment below](#issuecomment-2854717751): further discussion is needed on whether to keep or deprecate this class.
   - (May 7, 2025) Per further comments and AZ Digital discussion, we'll keep this class.
4. Should we go ahead and keep all the example classes on the Fonts page (.proxima-nova-bold, .proxima-nova-bold-italic, etc.)?
   - (May 7, 2025) I think we can keep these examples for now.

### How to Test
https://review.digital.arizona.edu/arizona-bootstrap/issue/1402/


mmunro-ltrr (commented):
Didn't we want to update the `font-family` for `<em>`, `<strong>`, &c. everywhere (not just for the font samples in the docs, but so that the actual Proxima Nova Italic, Bold, & so on would show up for normal uses like https://review.digital.arizona.edu/arizona-bootstrap/issue/1402/docs/5.0/content/typography/#inline-text-elements)?

bberndt-uaz:
Keep body color from upstream Bootstrap

bberndt-uaz:
> Is there a reason to have specific font named classes e.g. `.proxima-nova-bold-italic` vs using stackable classes e.g. `.bold` , `.italic` ?

Those classes are only included in the documentation as examples and we are considering whether to remove or update those.

bberndt-uaz:
Regarding today's AZ Digital discussion around the loading of fonts, I don't believe that we need to specifically set the `font-family` as "Proxima Nova Bold", "Proxima Nova Italic", and so on. The correct font is loaded by the browser thanks to the `font-face` declarations included in our stylesheet for Proxima Nova from Typekit, which you can view at https://use.typekit.net/emv3zbo.css.

For example, here is the `font-face` declaration for Proxima Nova Regular:
```css
@font-face {
font-family:"proxima-nova";
src:url("https://use.typekit.net/af/705e94/00000000000000003b9b3062/27/l?primer=7cdcb44be4a7db8877ffa5c0007b8dd865b3bbc383831fe2ea177f62257a9191&fvd=n4&v=3") format("woff2"),url("https://use.typekit.net/af/705e94/00000000000000003b9b3062/27/d?primer=7cdcb44be4a7db8877ffa5c0007b8dd865b3bbc383831fe2ea177f62257a9191&fvd=n4&v=3") format("woff"),url("https://use.typekit.net/af/705e94/00000000000000003b9b3062/27/a?primer=7cdcb44be4a7db8877ffa5c0007b8dd865b3bbc383831fe2ea177f62257a9191&fvd=n4&v=3") format("opentype");
font-display:auto;font-style:normal;font-weight:400;font-stretch:normal;
}
```

Note that three font formats are included for each font variation.

We might be concerned if the browser is applying "faux bold" or "faux italic" styling of its own rather than using the actual font variation files. More on that here:
 - https://fonts.google.com/knowledge/introducing_type/introducing_weights_styles#avoiding-fake-weights-and-styles
 - https://developer.mozilla.org/en-US/docs/Web/CSS/font-synthesis

One way we can confirm that this fake styling is NOT happening is by modifying the URLs in those `font-face` declarations and seeing what happens. Here is what we can see in Chrome for some bold text:

image
<p> </p>
Chrome DevTools indicates that the Proxima Nova Bold font is being loaded as a network resource.

As an experiment, let's see what happens when we override the URLs in that `font-face` rule for regular bold text with the URLs from the final `font-face` rule in the stylesheet:

image
<p> </p>

The font for bold text has been changed to use Proxima Nova Extra Condensed Medium Italics instead. Chrome shows that the font family is different even though I didn't change that in the `font-face` rule.

Related resources:
 - https://helpx.adobe.com/fonts/using/css-selectors.html
 - https://www.digitalocean.com/community/tutorials/how-to-load-and-use-custom-fonts-with-css
 - https://css-tricks.com/snippets/css/using-font-face-in-css/

danahertzberg:
Here are some of my thoughts.

> 1. Should we keep, deprecate, or remove `.heading-style` [from our custom styles](https://github.com/az-digital/arizona-bootstrap/blob/main/scss/custom/_styles.scss#L108-L112)?
>    - This class is not used in the Quickstart codebase. If we keep it, we should update it and rename it to `.az-heading-style`.
> 

I'm thinking we can safely remove it since we don't have a separate font family for headings anymore.

--
> 2. Should we keep, deprecate, or remove `.sans` [from our custom styles](https://github.com/az-digital/arizona-bootstrap/blob/main/scss/custom/_styles.scss#L108-L112)?
>    - This class assigns the same font-family we use for body and heading text. We use this class in one place in Quickstart, in the demo content (it can be easily removed). If we keep the class in Arizona Bootstrap 5, we should remove or separate out the font-weight setting.
>    - Quickstart includes a CKEditor "Sanserif Text" style which uses an incorrect class name, so it has no effect. I'll create a Quickstart issue to remove or update this style.

Thank you. Yes, let's remove sans since that is no longer used. Future updates may include adding a "serif" style to the style options in CKEditor.

--
> 3. Should we keep, deprecate, or remove `.text-hyphen` [from our custom styles](https://github.com/az-digital/arizona-bootstrap/blob/main/scss/custom/_styles.scss#L120-L123)?
>    - See https://digital.arizona.edu/arizona-bootstrap/docs/2.0/utilities/text/#hyphenate-word in the 2.0 docs and the [PR which added this class](https://github.com/az-digital/arizona-bootstrap/pull/310).
>    - This class is not used in Quickstart.

I'd like to see this fixed in Quickstart. The initial motivator was that headings within text on media would overflow the container on smallish screens. If this is fixed with BS5, no need to add on top of that. But I thought it should be working in QS. Let's review the motivating circumstances and see where we need to fix or if we can remove this class.

scss/override/_reboot.scss
```diff
@@ -0,0 +1,8 @@
+// Overrides to Reboot styles
+
+/*
+* > LINKS
+*/
+a {
+  font-weight: $font-weight-bold;
```
danahertzberg, line 7: This (I think) causes the sidebar menu items to be bold on the docs. Is there a way to leave the weight of the menu items in the docs the way they were?

bberndt-uaz: I have added a commit to restore the initial font-weight for sidebar links in our _custom-docs.scss file (where all customizations related to the docs should go).

joeparsons (dismissed):

mmunro-ltrr:
> Didn't we want to update the `font-family` for `\<em\>`, `\<strong\>`, &c. everywhere (not just for the font samples in the docs, but so that the actual Proxima Nova Italic, Bold, & so on would show up for normal uses like https://review.digital.arizona.edu/arizona-bootstrap/issue/1402/docs/5.0/content/typography/#inline-text-elements)?

This wasn't a real issue: I was mis-reading what the Inspector view of the page was showing, thinking the specific fonts were not getting used as they should in those cases. Subsequently confirmed by examining a magnified view of the rendered text.

mmunro-ltrr:
More on how good it would be to remove `.heading-style`: any additional styling should now use the CSS variables that upstream Bootstrap provides, at
https://github.com/twbs/bootstrap/blob/f849680d16a9695c9a6c9c062d6cff55ddcf071e/scss/_variables.scss#L652-L659
and there's no reason to override the default.

Similarly, there's no reason to keep `.sans`; upstream Bootstrap distinguishes sans & monospace/code fonts, but doesn't have a built-in serif option.

The hyphenation might be something we have to keep as a custom extension: I'm not seeing anything relevant in upstream Bootstrap 5.3.

mmunro-ltrr (dismissed):
The changes to the docs double as a good explanation of the complete PR (I should have read them first!).

mmunro-ltrr (approved):

joeparsons (approved):

trackleft (approved):

tadean (approved):

PR #1481, opened 2025-05-13
Use BS5 border radius mixins for custom nav styling

djcelaya:
Suggestion to consider using [BS5's default border-radius mixins](https://getbootstrap.com/docs/5.3/utilities/borders/#sass-mixins) instead of setting border-radius directly. I think I've used this one or twice with my buttons. It seems like this might be a slightly better way to customize border-radii in BS5.

https://review.digital.arizona.edu/arizona-bootstrap/issue/1403-djcelaya


bberndt-uaz (approved):
@djcelaya Thanks David!

PR #1507, opened 2025-05-19
Closes #1489: Make accordion backwards compatible

### LINKED ISSUE #1489: Make accordion backwards compatible
In #1478, we introduced an example of the old accordion implementation on the [backwards compatibility page](https://review.digital.arizona.edu/arizona-bootstrap/main/docs/5.0/backwards-compatibility/). Unfortunately, this example does not work with the new V5 accordion from TWBS5. I will attempt to make our old implementation compatible with AZBS5, or document the minimum number of breaking changes that must be addressed to use V5.

djcelaya:
https://review.digital.arizona.edu/arizona-bootstrap/issue/1489/

`data-` properties must be updated to `data-bs-` to allow for the collapse functionality to work. We could potentially consider this a breaking change.

There are additionally some styling differences. Should I try to change the deprecated accordion stylesheet to make the old implementation look like the new BS one? Or should we use the visual difference as a gentle reminder to update?

Where should this be placed on the Backwards Compatibility page? This seems like a "Deprecated Component Implementation" but not a "Deprecated Component". It will also likely be backwards incompatible


trackleft:
I've updated this to include the merge conflict fix from upstream.
The dist folder is now ignored by git, and only updated via CI build tools.

PR #1509, opened 2025-05-20
Closes #1508: Remove hard-coded 2.x reference in release workflow in preparation for creating 5.x releases.

### LINKED ISSUE #1508: Remove hard-coded 2.x reference in release workflow
The `release` GitHub Actions workflow currently contains a hard-coded reference to the `2.x` branch that will need to be removed before we can start creating 5.x releases.

joeparsons:
Removes hard-coded reference to 2.x in release workflow that would prevent 5.x releases from working.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,6 @@ jobs:
         uses: actions/checkout@v4
         with:
           token: ${{ secrets.REPO_DISPATCH_TOKEN }}
-          ref: 2.x
 
       - name: Set variables for Docker images
         run: |


mmunro-ltrr (approved):
I've not tested this on a replica of the project, but the defaults for `ref:` in the actions/checkout@v4 docs seem OK. @trackleft already set up https://github.com/az-digital/digital.arizona.edu to accept multiple branches when involved with the `peter-evans/repository-dispatch` workflow, but there's an open issue there https://github.com/az-digital/digital.arizona.edu/issues/42 which ideally needs resolved by not re-building the documents from scratch in that repository.

bberndt-uaz (approved):

PR #1550, opened 2025-05-29
Fix remaining Uncaught ReferenceError: bootstrap is not defined errors (targeting issue/1496)

joeparsons:
Still noticing `Uncaught ReferenceError: bootstrap is not defined` console errors when reviewing #1547 (specifically on https://review.digital.arizona.edu/arizona-bootstrap/issue/1496/docs/5.0/components/modal/).

We should be referencing our global `arizona-bootstrap` but `arizona-bootstrap` is not a valid name for an object reference (variable name) in JS so this updates the rollup config that controls that and some remaining references in the docs JS files.

### DIFF
--- a/build/rollup.config.mjs
+++ b/build/rollup.config.mjs
@@ -61,7 +61,7 @@ const rollupConfig = {
 }
 
 if (!ESM) {
-  rollupConfig.output.name = 'arizona-bootstrap'
+  rollupConfig.output.name = 'arizonaBootstrap'
 }
 
 export default rollupConfig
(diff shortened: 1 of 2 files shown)


joeparsons:
Rollup docs for the `output.name` configuration option (confirmed it is for a variable name so we don't want to use hyphens):
https://rollupjs.org/configuration-options/#output-name

build/rollup.config.mjs
```diff
@@ -61,7 +61,7 @@ const rollupConfig = {
 }
 
 if (!ESM) {
-  rollupConfig.output.name = 'arizona-bootstrap'
+  rollupConfig.output.name = 'arizonaBootstrap'
```
joeparsons, line 64: I guess another alternative to this would be to just use `bootstrap` here and not update the docs JS files...

Not sure how likely it is that custom site implementations of Arizona Bootstrap would need to reference this global object variable name...

bberndt-uaz (approved):

PR #1568, opened 2025-06-06
Closes #1567: Prevent photo gallery captions from overlapping images on mobile devices

### LINKED ISSUE #1567: Prevent photo gallery captions from overlapping images on mobile devices
## Motivation
The photo gallery custom component was updated for Arizona Bootstrap 5 with this pull request:
 - #1549

The changes in that PR should prevent the image caption element from overlapping the image itself on desktop. However, it is still possible for overlap to occur on mobile devices:

Image

Instead of fixing the caption to the bottom of the viewport and vertically centering the image, we could vertically center both the image and caption together.

The previous and next arrows could be changed to be vertically centered over each image, with their position adjusted according to the height and position of the image. However, having arrows that move up and down when switching images would introduce a usability issue, so it would be best to keep the arrows centered vertically based on the viewport.

bberndt-uaz:
This PR adjusts the display of photo gallery images and captions on mobile devices. Instead of having the caption fixed to the bottom of the viewport, this PR vertically centers the image and caption together. This prevents cases where a longer caption could cover part of an image with a portrait aspect ratio.

This PR also includes an improvement for the display of images with a landscape aspect ratio on desktop. The photo gallery demo images have been updated: the second image now serves as an example landscape image. 

### How to test
Compare the mobile display of the photo gallery component. For the third image, you can use Chrome DevTools to add a `<br>` to the caption on the main branch review site. 

This PR:
 - https://review.digital.arizona.edu/arizona-bootstrap/issue/1567/docs/5.0/components/photo-gallery/

Current display in the main branch:
 - https://review.digital.arizona.edu/arizona-bootstrap/main/docs/5.0/components/photo-gallery/

### Potential follow-up issues
 - For desktop, use JavaScript to fully center landscape aspect ratio images vertically in the carousel body
 - For mobile devices, use JavaScript to keep the caption element at the bottom and resize/center the image as necessary
 - Add demo of "Slider" style Photo Gallery


danahertzberg:
The image with the hand and blue background. It looks like it's still using the carousel crop; is that true? If so, can the carousel modal display the full image instead?

bberndt-uaz:
> The image with the hand and blue background. It looks like it's still using the carousel crop; is that true? If so, can the carousel modal display the full image instead?

I cropped that image file on purpose so that we would have an example of an image with a landscape aspect ratio. I think landscape images might be commonly used on our sites, so it would be good to have a demo image like that both in Arizona Bootstrap and in Quickstart.

mmunro-ltrr (approved):
Checked on a real mobile device as well as in a DevTools type of view.

danahertzberg:
> > The image with the hand and blue background. It looks like it's still using the carousel crop; is that true? If so, can the carousel modal display the full image instead?
> 
> I cropped that image file on purpose so that we would have an example of an image with a landscape aspect ratio. I think landscape images might be commonly used on our sites, so it would be good to have a demo image like that both in Arizona Bootstrap and in Quickstart.

Okay I just wanted to make sure the modal display was showing the full image. So if that is the full crop of the image outside of Drupal, then I think we're good.

tadean (approved):

PR #1636, opened 2025-07-15
Issue 1464 - Restore functionality of the Color Contrast docs page

zsaenz:
The interactive table on the custom Color Contrast page was not working in the latest version of our documentation. This used to be a JQUERY function loaded in the custom-docs file, where jQuery is no longer used, so we needed to swap it for a non-jQuery solution.

Didn't seem to be good idea to add this to custom-docs. Instead, added a file color-contrast.js called in application.js. I put it in /partials folder because it didn't seem to work elsewhere.

### DIFF
--- a/site/assets/js/partials/color-contrast.js
+++ b/site/assets/js/partials/color-contrast.js
@@ -5,11 +5,9 @@ if(toggleHideInaccessibleBox !== null) {
   toggleHideInaccessibleBox.addEventListener("change", (event) => {
     for(i = 0; i < inaccessibleElems.length; i++) {
       if(toggleHideInaccessibleBox.checked) {
-        inaccessibleElems[i].style.display = "none";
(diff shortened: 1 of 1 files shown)


camikazegreen:
Review site URL:
https://review.digital.arizona.edu/arizona-bootstrap/issue-1464

zsaenz:
Updated, take a look

mmunro-ltrr:
I'd a suggestion for a slightly different approach for making & including the updated .js file, but it was difficult to include in a series of comments here, so there's a new freestanding PR to show this at https://github.com/az-digital/arizona-bootstrap/pull/1637

mmunro-ltrr:
The alternative PR using `extra_js` in the front matter of the Markdown file `color-contrast.md` now correctly links the .js file: an initial slash was preventing the correct relative URL treatment there.

PR #1637, opened 2025-07-16
Closes #1464: An alternative version to updating the .js to toggle inaccessible color combinations in the docs.

### LINKED ISSUE #1464: Restore functionality of the Color Contrast docs page
The interactive table on the custom Color Contrast page is not working in the latest version of our documentation. 

## Conditions of satisfaction
- [ ] Update the content, styling, and functionality of the page for Arizona Bootstrap 5.

mmunro-ltrr:
There were a few suggested changes to https://github.com/az-digital/arizona-bootstrap/pull/1636, so this PR is a way to gather them together for easy reference, rather than as comments there.

People viewing the table of different text and background colors in the docs should have an interactive toggle, allowing them to turn on or off the display of the inaccessible combinations, but only at that one place. That's best addressed by having a .js file loaded just on the page containing the table, with a close analogy being the sample code for validating forms. The file for that is at `site/static/docs/5.0/assets/js/validate-forms.js`, and is included in the page that needs it by the `extra_js:` mechanism in `validation_md`, so we can make a new `toggle-visible.js` file beside that, then reference it in `color-contrast.md`.

### DIFF
--- a/site/assets/js/custom-docs.js
+++ b/site/assets/js/custom-docs.js
@@ -1,17 +1,6 @@
 (function ($) {
   'use strict'
 
-  // COLOR CONTRAST TABLE
-  $('#hide-inaccessible').change(function () {
-    if ($('#hide-inaccessible').is(':checked')) {
-      $('.inaccessible').css('visibility', 'hidden') // checked
-      $('.inaccessible').parent().css('opacity', '0')
-    } else {
-      $('.inaccessible').css('visibility', 'visible') // unchecked
-      $('.inaccessible').parent().css('opacity', '1')
-    }
-  })
-
   var specimenButtons = document.querySelectorAll('.js-specimen-modal-trigger');
   [].forEach.call(specimenButtons, function (specimenButton) {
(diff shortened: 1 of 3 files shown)


site/assets/js/custom-docs.js
```diff
@@ -1,17 +1,6 @@
 (function ($) {
```
trackleft, line 1:
```suggestion
(function () {
```

trackleft: Also remove Jquery from the last line

mmunro-ltrr:
The `extra_js` in the Markdown file works after all. There was a typo in the URL: the leading slash was preventing it being treated as relative.

zsaenz (approved):
#1636 
Can un-merge and close that pr

site/assets/js/custom-docs.js
joeparsons: Will be handled in #1639.

bberndt-uaz (approved):

tadean (approved):

PR #1656, opened 2025-07-31
Backport the upgrade to Node.js 22 from the main branch.

mmunro-ltrr:
This includes a cosmetic re-synchronization of the Dependabot configuration file with the main branch.


mmunro-ltrr:
Local builds replicate the failing build seen here, but with clearer error messages. One dependency is the legacy `node-sass` package, & that is not compatible with the latest LTS Node.js version.

joeparsons (approved):

PR #1711, opened 2025-08-29
Closes #1710: Add vertical spacing to page content for fixed-position Arizona Header on mobile

### LINKED ISSUE #1710: Add vertical spacing to page content for fixed-position Arizona Header on mobile
## Problem/Motivation
Why the issue was filed, steps to reproduce the problem, etc.

## Proposed resolution
Description of the proposed solution, the rationale behind it, and workarounds for people who cannot use the patch.

##  Screenshots
If applicable, add screenshots to help explain your problem. (Need a tool to share screenshots or screencasts as URLs? Try [monosnap](https://monosnap.com)

bberndt-uaz:

### DIFF
--- a/site/assets/scss/_custom-docs.scss
+++ b/site/assets/scss/_custom-docs.scss
@@ -192,6 +192,11 @@
   margin-right: 0;
   margin-left: 0;
 }
+@media screen and (max-width: 992px) {
+  body {
+    margin-top: unset;
+  }
+}
 
 // >> Background Wrappers
 .btn-background-wrapper-demo,
--- a/scss/custom/_arizona-header.scss
+++ b/scss/custom/_arizona-header.scss
@@ -9,6 +9,9 @@
 }
 
 @media screen and (max-width: 992px) {
+  body {
+    margin-top: 50px;
+  }
   .arizona-header {
     position: fixed;
     top: 0;


site/assets/scss/_custom-docs.scss
```diff
@@ -192,6 +192,14 @@
   margin-right: 0;
   margin-left: 0;
 }
+@media screen and (max-width: 992px) {
+  body {
+    margin-top: unset;
+  }
+  :target {
+    scroll-margin-top: revert;
```
bberndt-uaz, line 200:
```suggestion
    /* Restore upstream Bootstrap default. */
    scroll-margin-top: 80px;
```

mmunro-ltrr (dismissed):
(Just examined the demo site without building something using the component.)

djcelaya (dismissed):
Changes look good to me! I was testing locally with [this markup](https://github.com/user-attachments/files/22101823/index.html) (adjusting between 985-1000px) and the header no longer overlapped with the body content.

djcelaya:
> Does this only apply when using the ["extended" header option](https://digital.arizona.edu/arizona-bootstrap/v5/docs/5.0/components/arizona-header/#extending-the-header) with the menu? If the menu is not there, I would not expect the header to be sticky.

I observed locally that this styling applies whether or not the menu is present. Personally, I am not opposed to the Arizona Header always being present as that seems a little more visually consistent across our applications. I don't feel strongly about this opinion though and definitely see the argument where this is wasted space on mobile without a menu button.

danahertzberg:
> I observed locally that this styling applies whether or not the menu is present. Personally, I am not opposed to the Arizona Header always being present as that seems a little more visually consistent across our applications. I don't feel strongly about this opinion though and definitely see the argument where this is wasted space on mobile without a menu button.

I appreciate your opinion for sure. My opinion is pretty heavy on the side of not wanting to force a sticky header without the menu there. The shibboleth sign in page is an example where space is at a premium with announcements and form space. It also would be a largely noticeable change to the functionality on a page like that.

mmunro-ltrr:
I took a look at a local copy of the revised header & the effect if the CSS was included but the header itself was absent; the blank at the top did look a little strange. Would a potential follow-up be for the default behavior not to add any extra space, but have an extra class or property on the `<body>` to potentially enable it on specific pages?

site/content/docs/5.0/components/arizona-header.md
```diff
@@ -68,3 +78,22 @@ The Arizona Header can be extended with off-canvas search and navigation for mob
...
+
+
+## Disable fixed position on mobile devices
+
+In most cases, the Arizona Header element is intended to be fixed to the top of the browser window on mobile devices. This behavior is added by the `.az-fixed-header-on-mobile` class. This class can be omitted if a site does not include off-canvas navigation or other elements in the mobile display of the Arizona Header.
+
+**Note:** For this documentation site, this fixed-position behavior has been overridden to prevent the example Arizona Headers from disappearing behind the existing Arizona Bootstrap navbar above.
+
+{{< example >}}
+<div class="arizona-header bg-red"  id="header_arizona_example_2">
```
bberndt-uaz, line 90:
```suggestion
<div class="arizona-header bg-red" id="header_arizona_example_2">
```

tadean (approved):

PR #1806, opened 2025-11-20
Closes #1805: Add js-yaml as an explicit dependency to address CVE-2025-64718

### LINKED ISSUE #1805: Add js-yaml as an explicit dependency
Upstream Bootstrap has an explicit `js-yaml` dependency in `package.json`, but Arizona Bootstrap doesn't have it there, just as a transitive dependency which appears in `package-lock.json`. We didn't include the explicit dependency because it appeared around the time of the upstream switch from Hugo to Astro (so not relevant here yet), but Dependabot might be flagging this not just as an update, but as a security fix, and in any case we should just add the correct version to `package.json` (with a manual PR), & Dependabot should then automatically close this one.

_Originally posted by @mmunro-ltrr in https://github.com/az-digital/arizona-bootstrap/issues/1799#issuecomment-3555104119_

djcelaya:
Declares js-yaml 4.1.1 as an explicit devDependcy.

This declaration exists [upstream](https://github.com/twbs/bootstrap/blob/61b0bab5c29415938038a71592b30b05eb96e452/package.json#L151) already and corresponds with the version bump in #1799.

Review site: https://review.digital.arizona.edu/arizona-bootstrap/issue/1805/

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
     "globby": "^15.0.0",
     "hugo-bin": "^0.149.2",
     "jquery": "3.7.1",
+    "js-yaml": "^4.1.1",
     "linkinator": "6.1.2",
     "nodemon": "^3.1.10",
     "npm-run-all2": "^8.0.4",


joeparsons:
## Suggested release notes

```
Updated js-yaml NPM development dependency to version 4.1.1 which addresses [CVE-2025-64718](https://github.com/nodeca/js-yaml/security/advisories/GHSA-mh29-5h37-fv8m). Note: Sites using Arizona Bootstrap are not directly affected by this security advisory and its associated vulnerability. js-yaml is a development dependency.
```

PR #1842, opened 2025-12-19
Fixes #1777: Icons used in CSS pseudo element is read by screen readers

### LINKED ISSUE #1777: Icons used in CSS pseudo element is read by screen readers
## Problem/Motivation
Decrotive icons used in CSS pseudo-content is read by screen readers.

### Describe the bug
When icons are used in CSS pseudo-content, screen readers will read whatever is set in the content. Since the icons are often decrotive, this might cause confusion.
This issue was first mentioned (Oct 22, 2025) by Jeff at DRC when reviewing a NetID page.

### To Reproduce
Using Jaws on any platform/browser (also on Mac Voice Over) read the breadcrumb content for a page that is at least two levels deep and is showing a chevron in the breadcrumb.

## Proposed resolution
Some possible options:
- use an inline image E.g. <svg> or 
- use an encoded svg as the CSS pseudo-content with currentColor to inherit the font color
- use clip-path in the CSS with: currentColor to inherit the font color
- use an SVG for the background (not ideal)
[I have some examples from long ago](https://codepen.io/itsmemario/pen/WNObzwq)

### Expected behavior
It should be optional to have the icon read, or not be read at all because they often don't provide value.

##  Screenshots
[screenshot](https://limewire.com/d/v3ls5#Hm6UnkmtZd)

djcelaya:
Replaces the names of the Material Symbols Rounded icons used in the Breadcrumb with their respective Unicode codepoints, which screen readers do not read aloud.

### How to test
1. Navigate to the [Breadcrumb component documentation page](https://review.digital.arizona.edu/arizona-bootstrap/bug/1777/docs/5.0/components/breadcrumb/) at `/docs/5.0/components/breadcrumb/`.
2. Using a screen reader, confirm that neither the home icon or the chevron_right icon are read aloud by the screen reader.


joeparsons:
Do we need to make similar change to this icon used in Accordions?
https://github.com/az-digital/arizona-bootstrap/blob/b57d3d3511afcfe41bf58c9a745855b50e59f9cd/scss/deprecated/_accordion.scss#L39

tadean (approved):

joeparsons:
There is some discussion about code points vs icon names and screen readers in this old PR (you have to expand some of the collapsed history to find it):
https://github.com/az-digital/arizona-bootstrap/pull/532

(Seems we intentionally changed from code points to icon names to make them readable by screen readers at one point).

We should probably check with @accesswatch about best practices for icons within breadcrumb trails related to screen readers.

djcelaya:
From @accesswatch on Jan 7:
> Breadcrumb icons being announced is definitely not desired behavior, and I agree this needed correction.
> Changing the pseudo-content reference from a human-readable glyph name to a codepoint is a reasonable and effective improvement, and it’s encouraging to see that VoiceOver no longer announces the icon. JAWS and NVDA are also no longer announcing them either with this change.
>  
> Conceptually, this aligns with the expectation that breadcrumb separators are decorative and should not be exposed to assistive technology.
> My only recommendation before broad adoption is to ensure we are treating these icons as explicitly decorative wherever possible—either by keeping them fully within CSS pseudo-content or by applying aria-hidden="true" if a DOM element is involved—so that we’re not relying solely on screen-reader interpretation of codepoints.
> Assuming cross-screen-reader testing (particularly NVDA and JAWS) confirms the same behavior (and I did), I’m comfortable with this approach moving forward and would support including it in a future release but would encourage you to consider the hidden approach.

I think we're in good shape with this. Both of these icons are fully captured in pseudo-content and are decorative.

trackleft (approved):

PR #1848, opened 2026-01-05
Fixes #1827: Secondary navigation dropdown menu should expand on :hover or :focus

### LINKED ISSUE #1827: Secondary navigation dropdown menu should expand on :hover or :focus
The current implementation requires the user to click on the dropdown-toggle to expand the submenu. MarCom would like it to expand on hover instead. To maintain accessibility compliance, the submenu should also expand on `:focus`, which is gained via keyboard navigation.
Image

## Conditions of satisfaction
- [x] When a user hovers over a dropdown-toggle at the primary level, the corresponding secondary level dropdown menu should be shown.
- [x] When a dropdown-toggle gains `:focus` via keyboard navigation at the primary level, the corresponding secondary level dropdown menu should be shown.

djcelaya:
~CSS-only fix to allow a secondary level dropdown menu to be shown upon hover or focus of the corresponding primary level dropdown toggle.~

CSS and JS fix to allow the BS5 Dropdown component to handle hover events. A CSS-only fix is insufficient because ARIA  properties in the HTML need to be dynamically updated to support accessibility. This approach attempts to extend the BS5 Dropdown JS component so that BS5 is handling the show and hide events of the dropdown-menu. I was previously doing this with CSS which bypasses some BS5 JS that correctly handles ARIA.

### How to test
1. Navigate to the [AZ Navbar](https://review.digital.arizona.edu/arizona-bootstrap/issue/1827/docs/5.0/components/navbar/#az-navbar) documentation section at `/docs/5.0/components/navbar/` OR the [AZ Navbar Example](https://review.digital.arizona.edu/arizona-bootstrap/issue/1827/docs/5.0/examples/navbar-az/) page at `/docs/5.0/examples/navbar-az/`.
2. Hover your cursor over the Single Dropdown navigation element. The corresponding secondary level dropdown menu should appear when you hover over either the "Single Dropdown" text or the the down chevron toggle button. If needed, compare with the [Live](https://digital.arizona.edu/arizona-bootstrap/v5/docs/5.0/components/navbar/#az-navbar) environment.
3. Hover your cursor over the Split Dropdown navigation element. The corresponding secondary level dropdown menu should appear only when you hover over the down chevron toggle button. Hovering over the "Split Dropdown" text should not auto-expand the submenu as it will likely be pointing to an actual page. If needed, compare with the [Live](https://digital.arizona.edu/arizona-bootstrap/v5/docs/5.0/components/navbar/#az-navbar) environment.

### DIFF
--- a/js/index.esm.js
+++ b/js/index.esm.js
@@ -19,7 +19,7 @@ export { default as Toast } from '../node_modules/bootstrap/js/src/toast.js'
 export { default as Tooltip } from '../node_modules/bootstrap/js/src/tooltip.js'
 export { default as fixModalAriaHidden } from './src/modal.js'
 export { default as photoGalleryGridSlideToImage } from './src/photogallery.js'
-export { default as enableAzNavbarHoverDropdowns } from './src/navbar.js'
+export { default as enableAzNavbarHoverDropdowns } from './src/navbar-hover-dropdown.js'
 
 /**
  * Temporary fix for blocked aria-hidden attribute on modals.
--- a/js/index.umd.js
+++ b/js/index.umd.js
@@ -19,7 +19,7 @@ import Toast from '../node_modules/bootstrap/js/src/toast.js'
 import Tooltip from '../node_modules/bootstrap/js/src/tooltip.js'
 import fixModalAriaHidden from './src/modal.js'
 import photoGalleryGridSlideToImage from './src/photogallery.js'
-import enableAzNavbarHoverDropdowns from './src/navbar.js'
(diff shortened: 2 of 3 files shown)


djcelaya:
> A couple focus style issues
> 
> * Change all border colors to default red (inset?)
> * Single dropdown style should not have an underline when menu is already open (focus style)

Follow up issues have been created. See #1855 and #1854.

mmunro-ltrr (dismissed):

joeparsons (approved):

PR #1918, opened 2026-03-03
Closes #1894: Use CSS-embedded SVG for wordmark logo in Arizona Header component

### LINKED ISSUE #1894: Use inline SVG or packaged image file for wordmark logo in Arizona Header component
## Motivation
The Arizona Header component code examples use an absolute cdn.digital.arizona.edu URL for the wordmark line logo image file (SVG).

This means that websites using the Arizona Header component as it appears in the Arizona Bootstrap documentation are always referencing an image file on cdn.digital.arizona.edu. 

This proved problematic recently when cdn.digital.arizona.edu was impacted by an AWS CloudFront outage.  Some AZ Quickstart site owners changed their Arizona Barrio theme settings to use a local copy of Arizona Bootstrap rather than the CDN-hosted version to mitigate the outage but found that the wordmark logo image URL was still hard-coded to the CDN.

## Proposed Resolution
We should consider replacing the `` element containing the cdn.arizona.edu `src` attribute with an inline `<svg>` element or package the SVG file with Arizona Bootstrap somehow so that local copies of the library can use the component without relying on cdn.digital.arizona.edu if desired.

djcelaya:
This is a **CSS-embedded SVG** approach to #1894. It defines the wordmark as a CSS variable in the `.arizona-header` scope. If we legitimately anticipate making changes to the Arizona header custom component in the future (i.e. changing background color to blue), this approach is perhaps preferred to #1919. It gives us the most control at the CSS level and does not require a discrete image asset to be added or maintained by the project.

This approach is technically different than the proposed inline SVG implementation. That would require the `<svg></svg>` in its entirety to be included in the HTML. And although possible, the contents of https://cdn.digital.arizona.edu/logos/v1.0.0/ua_wordmark_line_logo_white_rgb.min.svg is enormous and would be difficult to use on the HTML side.

This approach also no longer requires an `` element to be used since there is no image file to reference. I am currently using `<span>`s in place of the `` tag. The same CSS class `.arizona-line-logo` is used but it is updated to set an explicit size on the `<span>`. `` elements set their size based on their src but `<span>` elements with background images do not have any such mechanism so we must set the size explicitly.

Since this changes the required HTML structure of the Arizona header component and because it redefines an already used class to support that new structure, I would recommend this as a minor release change at a minimum.

### How to test
1. Navigate to the [Arizona header](https://review.digital.arizona.edu/arizona-bootstrap/issue/1894-inline-svg/docs/5.1/components/arizona-header/) component page.
2. Compare with [LIVE](https://digital.arizona.edu/arizona-bootstrap/v5/docs/5.1/components/arizona-header/) version. Visually, these should be the same.

Review site: https://review.digital.arizona.edu/arizona-bootstrap/issue/1894-inline-svg/

### Questions
1. Is the keyboard experience of this implementation equivalent to the current experience?


mmunro-ltrr (approved):
Preferring this approach over https://github.com/az-digital/arizona-bootstrap/pull/1919

djcelaya:
Closing for now. This is not a current priority. Can be resurrected at a later date if desired.

Issue #1994, opened 2026-04-24
Add accessibility regression checks and manual verification guidance


accesswatch:
## Problem
The repository currently builds and lints docs, but it does not appear to run dedicated accessibility tooling or publish explicit manual verification steps for the most fragile interactions.

## Impact
- Regressions in docs examples and custom interactions are easy to miss.
- Hover/focus and validation behavior require manual validation but are not tracked as part of the workflow.

## Source
- `package.json`
- Custom interactions in `js/src/navbar.js`
- Validation script in `site/static/docs/5.1/assets/js/validate-forms.js`

## Acceptance Criteria
- The repo has a documented accessibility verification path for docs/example changes.
- At least one automated accessibility check is evaluated or introduced for the docs site.
- Manual verification guidance covers keyboard and screen reader testing for the highest-risk components.

Issue #1993, opened 2026-04-24
Clean up docs shell focus and skip-link behavior


accesswatch:
## Problem
The docs layout includes a focusable input inside an `aria-hidden` container, and the global skip link lands on the page heading instead of the main landmark.

## Impact
- Focus could land in content hidden from assistive technologies.
- Repeated-content bypass is less robust than it should be.

## Source
- `site/layouts/_default/docs.html`
- `site/layouts/_default/about.html`
- `site/layouts/partials/skippy.html`

## Acceptance Criteria
- No focusable element remains inside `aria-hidden` content.
- The skip link targets the main landmark directly.
- Docs and about layouts use a consistent skip-target strategy.

Issue #1992, opened 2026-04-24
Revise inaccessible validation guidance and checkout demo patterns


accesswatch:
## Problem
The validation docs explicitly acknowledge that the client-side custom validation patterns are not accessible, but the page still publishes and promotes those patterns. The checkout demo also includes an unlabeled promo code input.

## Impact
- Users copy inaccessible validation behavior from the docs.
- Error messaging is not robustly associated or announced.
- The checkout demo includes an unlabeled field.

## Source
- `site/content/docs/5.1/forms/validation.md`
- `site/content/docs/5.1/examples/checkout/index.html`
- `site/static/docs/5.1/assets/js/validate-forms.js`

## Acceptance Criteria
- The docs clearly steer users to accessible validation patterns.
- Published examples associate errors with fields and define a focus-management strategy for invalid submission.
- The checkout promo code field has a programmatic label.
- Any remaining inaccessible patterns are clearly contained as legacy/reference content rather than recommended guidance.

Issue #1991, opened 2026-04-24
Align docs shell branding labels and disclose new-tab behavior


accesswatch:
## Problem
The docs shell still labels Arizona Bootstrap brand links as Bootstrap, and several docs/footer links open new tabs without any visible or assistive-technology cue.

## Impact
- Brand naming is inconsistent for screen reader users.
- New-tab navigation creates avoidable context changes.

## Source
- `site/layouts/partials/docs-navbar.html`
- `site/layouts/partials/footer.html`
- `site/content/docs/5.1/about/get-involved.md`

## Acceptance Criteria
- Brand links expose Arizona Bootstrap as the accessible name.
- New-tab links are either disclosed consistently or revised to avoid forced new-tab behavior.
- Docs-shell wording is consistent across navbar, footer, and supporting content.

PR #1995, opened 2026-04-24
docs: add accessibility review planning artifact

accesswatch:
## Summary
Add the April 24, 2026 accessibility review and remediation planning document to the repository.

## What this PR includes
- Adds `ACCESSIBILITY-REVIEW-PLAN-2026-04-24.md`
- Captures the accessibility findings from the repository review
- Documents the remediation sequence and residual risks
- Links the GitHub planning epic and sub-issues created from the review

## Why
This creates a reviewable planning artifact in the repository before implementation work starts, so the team can align on scope, priorities, and issue breakdown.

## Related work
- Related to #1987
- Planning issues: #1988, #1989, #1990, #1991, #1992, #1993, #1994

### DIFF
--- a/ACCESSIBILITY-REVIEW-PLAN-2026-04-24.md
+++ b/ACCESSIBILITY-REVIEW-PLAN-2026-04-24.md
@@ -1,0 +1,273 @@
+# Accessibility Review And Remediation Plan
+
+## Review Metadata
+
+- Repository: Arizona Bootstrap
+- Review date: April 24, 2026
+- Planning branch: `accessibility/planning-2026-04`
+- Parent issue: [#1987 - Epic: remediate April 2026 accessibility review findings](https://github.com/az-digital/arizona-bootstrap/issues/1987)
+
+## Setup Completed For Review
+
+The repository needed local setup before the docs site could be reviewed in rendered form.
+
+- Installed workspace dependencies with `npm install`
+- Installed `jq` with `winget`
+- Exposed Git for Windows `sh.exe` on the user `PATH`
+- Generated `hugo.yml` from `hugo_template.yml` using the existing repository script
+- Confirmed the docs site builds successfully with `npm run docs-build`
+
+## Executive Summary
+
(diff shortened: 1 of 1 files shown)


djcelaya (changes requested):
@accesswatch, is it correct that the majority of this information is captured on #1987 and its linked sub-issues? If so, do you think we could close this PR and just maintain the planning in the issue space? It seems a little off to me to commit a report to the repo.

accesswatch:
Hi David, thanks for the thoughtful feedback. I see your point about whether planning documents should live in the repo. I think there's value in having a permanent, version-controlled record of our accessibility planning process, especially as we're trying to establish better accessibility practices. However, I take your point about discoverability. What if we kept the high-level planning and ongoing discussions in the issues (as you suggested), but committed a streamlined version of the final planning document to the repo for historical reference? This way we get the benefits of both approaches - collaborative planning in issues and a permanent record in the repo. I'm open to other suggestions though - what do you think would work best for our team's workflow?

djcelaya:
> Hi David, thanks for the thoughtful feedback. I see your point about whether planning documents should live in the repo. I think there's value in having a permanent, version-controlled record of our accessibility planning process, especially as we're trying to establish better accessibility practices. However, I take your point about discoverability. What if we kept the high-level planning and ongoing discussions in the issues (as you suggested), but committed a streamlined version of the final planning document to the repo for historical reference? This way we get the benefits of both approaches - collaborative planning in issues and a permanent record in the repo. I'm open to other suggestions though - what do you think would work best for our team's workflow?

I suppose a streamlined document would be preferred. It still seems fairly redundant to me TBH. Will we update this document in the future or do we plan on adding more of these? Are we doing this in the Quickstart space as well?

I am fine with moving ahead with the full report or a condensed report depending on what others feel. I think my preference would still be to maintain these reports and related planning artifacts in the GitHub space (i.e. issues, PRs, projects, etc) and not the .git space (in the repo).

Issue #2052, opened 2026-05-29
Finalize AZ Navbar Fullscreen


djcelaya:
Continuation of #1999.

## Mockups
- [Arizona.edu Redesign - Dev](https://xd.adobe.com/view/f28d87c5-418d-4c27-b7cc-627e99ca2c14-a7de/grid/?hints=off) (current)
- [Arizona.edu Navigation Dev](https://xd.adobe.com/view/72056437-64ab-4b34-9c61-ade0cf25b6c6-2756/grid/) (original)
  - [Vertically constrained](https://xd.adobe.com/view/72056437-64ab-4b34-9c61-ade0cf25b6c6-2756/screen/0ec31c57-5325-4591-9258-c52950a2426f/) - demonstrates independently scrollable nav columns
- [Arizona.edu Mobile Navigation Dev](https://xd.adobe.com/view/534efc74-d9f1-4c48-b286-815f5e95f08a-061d/grid)
- [Desktop header](https://xd.adobe.com/view/5e5859af-459f-438c-83e0-963b6791dd91-0b0b/) from [Arizona.edu Components](https://xd.adobe.com/view/5e5859af-459f-438c-83e0-963b6791dd91-0b0b/grid)
- [Mobile header](https://xd.adobe.com/view/5e5859af-459f-438c-83e0-963b6791dd91-0b0b/screen/2ccaceee-8512-4593-bda1-658172175159/) from [Arizona.edu Components](https://xd.adobe.com/view/5e5859af-459f-438c-83e0-963b6791dd91-0b0b/grid)
- [Mobile nav fixes](https://xd.adobe.com/view/c28674da-2d8e-4078-8e41-361137da3d4e-97d5/specs/)

## Conditions of Satisfaction
- [x] Follow up with @mmunro-ltrr about [this former review comment](https://github.com/az-digital/arizona-bootstrap/pull/2016#pullrequestreview-4392013253). See #2113.

djcelaya:
In the BERB meeting on 6/2, we decided that the wordmark will not be configurable at this time. @bberndt-uaz, this suggests to me that the wordmark should continue to be included with BS5 and used as the default at the Bootstrap level. Do you agree?

camikazegreen:
[...]

Issue #2075, opened 2026-06-02
AZ Navbar Fullscreen (mobile): Refactor mobile nav JavaScript


bberndt-uaz:
The JavaScript required for the AZ Navbar Fullscreen mobile display could use further refactoring.

## Conditions of satisfaction
- [x] Confirm if there could be security issues with innerHTML. If necessary, sanitize strings and/or change how elements are updated.
- [x] Confirm if event listeners on the mobile column need to be removed. Investigate event delegation.
- [x] Save DOM queries to be reused.
- [x] Create constants for additional strings.
- [x] Other general refactoring to improve maintainability and performance.

PR #2101, opened 2026-06-11
Closes #2085, #2090, #2110: AZ Navbar Fullscreen: Sync desktop nav column heights, refine scroll fade, and improve scrollbar styling

### LINKED ISSUE #2085: AZ Navbar Fullscreen: Check scrollbar styling in non-Safari browsers
Non-Safari browsers are displaying the scrollbars that appear when the viewport is vertically condensed with a gray color–which clashes with the design. Let's see if we can update it to a blue color.

Image

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/scrollbar-color

Safari's default styling of these scrollbars is not a bad starting point:

Image

### LINKED ISSUE #2090: AZ Navbar Fullscreen: Investigate if nav column heights can be reduced to max content-height of all visible nav columns
The modal menu of AZ Navbar Fullscreen presents the main menu with a three column layout. The left column presents the primary navigation items, the middle columns shows the secondary navigation tier, and the right column holds the third navigation level. The height of each of these columns is determined by the number of navigation items they contain. The [design mockup](https://xd.adobe.com/view/f28d87c5-418d-4c27-b7cc-627e99ca2c14-a7de/grid/?hints=off) requires that there a be a vertical separator between each of the three columns–so two in total. The mockup also suggests that the height of these separators be equal. Additionally, each separator should be at least as tall as the tallest navigation column that it separates. This creates the overall effect that the two separators are always as tall as the tallest navigation column currently displayed. If the displayed navigation structure changes as a result of user input, the height of the separator should be updated to match the changed navigation. See the following two screenshots from the mockups that illustrate the dynamic height of the separators. It is highly desired that this be resolved with a HTML- and CSS- solution, JavaScript should only be used as a last resort.

Image

In the above example, the height of these two column separators is determined by the content height of the primary nav column.

Image

In this second example, the height is now determined by the secondary nav column since its increased menu has caused its content height to exceed that of the primary nav column. Although this is only a slight visual change drawn from the mockups, all three navigation levels are unbounded in the number of menu items they could contain

When the viewport is [vertically constrained](https://xd.adobe.com/view/72056437-64ab-4b34-9c61-ade0cf25b6c6-2756/screen/0ec31c57-5325-4591-9258-c52950a2426f/), the modal header and footers should remained fixed to the top and bottom of the viewport, respectively. When the modal body that holds the modal menu loses height, the three navigation columns should become vertically scrollable and they should be independently scrollable. That is, scrolling the first tier should not scroll the second tier.

Additionally, a "fade under" effect is desired when any of the three nav columns independently scroll out of view. With the fixed modal header and footers, they effectively fade under these two areas.

Image

Image

If possible, it is highly desirable that the current keyboard and tab navigation experiences be maintained (specifically without JS). Adjusting the HTML structure or SCSS is preferred. If an HTML adjustment is needed that breaks either experience, an alternative solution should be presented.

### LINKED ISSUE #2110: AZ Navbar Fullscreen: Scrollable nav column does not allow final item to clear fade overlay
Image

There should be some bottom spacing on the last item in each nav column so that, if the viewport is vertically constrained, there is some point where the last item is totally unobscured by the fade effect.

djcelaya:
See #2085, #2090 and #2110.

## Summary

This PR improves the desktop AZ Fullscreen Navbar modal menu behavior and presentation by:

- **Desktop Column Synchronization** (#2090): Synchronizing visible nav column heights to the tallest visible nav content while preserving per-column scrolling when content exceeds available viewport height
- **Scroll Fade Refinement** (#2110): Moving top fade visual effect from individual scrollable columns to the modal header, ensuring content clears both top and bottom fade zones without obstruction; exposing fade sizing as CSS custom properties for runtime control
- **Scrollbar Styling** (#2085): Styling scrollbars in non-Safari browsers to use a blue color palette aligned with the design system instead of the default gray
- **Documentation & Examples**: Updating fullscreen example markup/classes for primary/secondary/tertiary nav consistency and documenting the desktop column synchronization behavior with required component classes

## Implementation Details

- Added desktop nav column height synchronization module and explicit primary column targeting:
  - js/src/navbar-az-fullscreen.js
- Wired fullscreen synchronization into library entry points:
  - js/index.esm.js
  - js/index.umd.js
- Updated fullscreen menu styles with improved fade handling, scrollbar styling, and separator behavior:
  - scss/custom/_navbar-fullscreen.scss
- Updated fullscreen example pages for structure and state coverage:
  - site/content/docs/5.1/examples/navbar-az-fullscreen/*
- Updated fullscreen component documentation with desktop synchronization subsection and example markup alignment:
  - site/content/docs/5.1/components/navbar.md

## How To Test

Review site: https://review.digital.arizona.edu/arizona-bootstrap/issue/2090/
See #2052 for mockups.

1. Open the [AZ Navbar Fullscreen example pages](https://review.digital.arizona.edu/arizona-bootstrap/issue/2090/docs/5.1/examples/navbar-az-fullscreen/) and launch the modal menu on desktop (lg+).
2. Expand/collapse primary and secondary menus to show 1/2/3 visible columns.
3. Verify visible columns are equal height to the tallest visible nav content.
4. Verify scrollbars still appear per column when content exceeds available height.
5. Verify top fade effect is positioned below the header and last nav items can scroll into view without being obscured.
6. Test in Firefox and Chrome to verify scrollbars render with blue styling.


scss/custom/_navbar-fullscreen.scss
```diff
@@ -589,20 +574,68 @@
     --bs-nav-link-hover-color: var(--az-navbar-fullscreen-modal-menu-hover-color);
     flex-direction: column;
     flex-wrap: nowrap;
-    margin-right: var(--az-navbar-fullscreen-modal-menu-col-spacing-x);
-    margin-left: var(--az-navbar-fullscreen-modal-menu-col-spacing-x);
+    // margin-right: var(--az-navbar-fullscreen-modal-menu-col-spacing-x);
```
bberndt-uaz: I believe this margin-right property is still needed?

#### Before:

Image

#### After:

Image

bberndt-uaz:
Since you are introducing an actual navbar-az-fullscreen.js, would you mind fixing the comment line below to say `navbar-az-fullscreen-mobile-nav.js` as part of this PR? I was intending to fix this as part of #2075, but it would be good to get it fixed here.

https://github.com/az-digital/arizona-bootstrap/blob/d901a192bc5a18041075ee1d807b34bfe6d9795b/js/src/navbar-az-fullscreen-mobile-nav.js#L3

scss/custom/_navbar-fullscreen.scss
```diff
@@ -661,9 +698,18 @@
...
     margin-left: 0;
+
     font: var(--az-navbar-fullscreen-modal-menu-primary-font);
+    // border-right: 1px solid var(--bs-azurite);
     letter-spacing: .1px;
+    border-left: 0;
+
+    > .nav-item > .nav-toggle {
+      --az-navbar-fullscreen-nav-toggle-size: 40px;
+      --az-navbar-fullscreen-nav-toggle-separator-offset: 24px;
```
bberndt-uaz: I don't believe this change is necessary? It seems to be causing a regression with the nav toggle buttons for the mobile display:

#### Before:

Image

#### After:

Image

```suggestion
```

djcelaya:
> Since you are introducing an actual navbar-az-fullscreen.js, would you mind fixing the comment line below to say `navbar-az-fullscreen-mobile-nav.js` as part of this PR? I was intending to fix this as part of #2075, but it would be good to get it fixed here.
> 
> https://github.com/az-digital/arizona-bootstrap/blob/d901a192bc5a18041075ee1d807b34bfe6d9795b/js/src/navbar-az-fullscreen-mobile-nav.js#L3

Fixed with https://github.com/az-digital/arizona-bootstrap/pull/2101/changes/1efa99dd0847c045df5d9c2615d2a9f031c09726. Thanks!

scss/custom/_navbar-fullscreen.scss
djcelaya: Good catch. Removed in https://github.com/az-digital/arizona-bootstrap/pull/2101/changes/8d9f3219fef3aefade9fa4c1dc45a93a82afbd0e. Thanks!

scss/custom/_navbar-fullscreen.scss
djcelaya: Thanks for looking at this, @bberndt-uaz! I am having trouble reproducing the After screenshot above. The intent is for the .nav-items to now have 8px right padding + ~16px for the stable scrollbar = 24px between column nav-items and the column separator.

I have tried in Safari, Firefox, and Chrome but can't get the scrollbar to touch the nav-items like the Above screenshot suggests. Could you provide any more details about how you saw this?

As an aside, I will focus more on exact scrollbar styling in (the now outdated) #2107.

mmunro-ltrr: I was just able to replicate the screenshot showing the buttons touching the scrollbar (but couldn't persuade my screen capture utility to include the scrollbar on the image). Firefox 147.0.1 on Linux.

djcelaya: Thanks for the confirmation, @mmunro-ltrr. I will dig in a bit more and see what I can find.

djcelaya: I've been tweaking some things in https://github.com/az-digital/arizona-bootstrap/pull/2101/changes/cd49727b0b1768411edcbea0ea47daabe29eeca0. When either of you have time, could you check again and see if this is working any better? This is what I'm seeing in my Firefox so if you're still seeing the scrollbar touch the toggle button, any additional details you can think of would be helpful.

Screenshot 2026-06-18 at 10 39 58 AM

mmunro-ltrr: Screenshot_2026-06-18_13-44-29
An attempt to replicate the latest screenshot. This particular Firefox version seems to render the scrollbars differently, so the dark bars fade in & out on a mouseover, and they seem too narrow relative to the other screenshot, but resizing the window doesn't cause overlaps with the buttons.

djcelaya: Thanks! This is helpful. I'm running Firefox 152.0.1 on macOS so that may be contributing here. If I had to guess from your screenshot, it looks like the navs have like maybe 8px of right spacing and no space is reserved for the scrollbar.

djcelaya: I'm not sure if this is better than the direction I've been trying to pursue on this PR, but here is an alternative approach to handle scrollbar "sizing": https://github.com/az-digital/arizona-bootstrap/pull/2115

mmunro-ltrr: I've re-done the screenshot on https://github.com/az-digital/arizona-bootstrap/pull/2115 (so will post that there for comparison).

bberndt-uaz: I just checked the updated review site in Chrome and Firefox and with the normal window I don't see the margin issue. However, with DevTools / Responsive Design Mode open and set to use a 1440p screen, I do still see the issue for some reason:

image

djcelaya: Shoot! Scrollbars are becoming my new enemy. Thanks for looking into it!

mmunro-ltrr: Looked again at this (and #2115 ) on a real Samsung S20+ with Firefox 151.0.4 on Android 13, & wasn't seeing the problem.

djcelaya: I was playing around with this again this morning. When you have a moment, @mmunro-ltrr, how is this most recent iteration looking?

mmunro-ltrr: It still looks fine with Firefox on a real phone, and with Chromium on Linux
Screenshot_2026-06-22_16-35-53
but the width of the scrollbars themselves seems too narrow in Firefox (& they don't appear in the screenshot because a hover is needed to reveal them)
Screenshot_2026-06-22_16-36-22
Does it count as more of a browser bug than anything else? It might be good to get the Browserstack comparisons working again.

mmunro-ltrr: Just confirmed the general appearance of the Firefox rendering with version 152.0.1 on macOS Tahoe 26.5.1.

djcelaya: Thanks @mmunro-ltrr! This is again very helpful. I think the problem is classical scrollbars vs the more modern overlay scrollbars. Classic scrollbars always take up space. Overlay scrollbars (more of a mobile default it seems) do not take up space and render over the content only when scrolling. When at rest, the scrollbar fades away so the content is not blocked. Unfortunately, it seems like there is not a way to request one scrollbar type vs the other, it is platform dependent. There also is apparently no way to set the width of either scrollbar type consistently.

To that end, I have now been able to reproduce this using Chrome DevTools like @bberndt-uaz described.

djcelaya: Given the hidden complexity of this scrolling issue (and the fact that no scrolling styles are defined in the mockups), I am electing to move forward with the simpler scroll approach demonstrated in #2115. I will review with Tom and if adjustments are needed, they will be handled in a follow up.

When vertically constrained, the approach on this PR should be about equivalent to what has already been deployed: https://live-azs-arizona.pantheonsite.io (plus the scrollbar coloring - which does come across reliably across browsers).

The fade overlay refinements and column separator height sync also present on this PR are important improvements so I would like to merge this soon if there are no other major issues.

mmunro-ltrr:
Looking better, & it was good to see the explicit `scrollbar-width: thin` go, even if it wasn't changing much.
Screenshot_2026-06-23_11-30-18

mmunro-ltrr (approved):

trackleft (approved):

danahertzberg (approved):

PR #2113, opened 2026-06-17
Closes #1698: derive the SVG files in `dist/css/img` from the originals in `scss/img`.

### LINKED ISSUE #1698: Restore handling of distributed SVG files
## Problem/Motivation
Changes to the SVG files used by the background wrapper custom component do not appear in the `dist/css/img` directory. This is *not* currently a release-blocking bug, because the failures are invisible for the current state of either branch of Arizona Bootstrap.

### Describe the bug
The PR https://github.com/az-digital/arizona-bootstrap/pull/157 moved the canonical versions of the SVG files that the [Background Wrappers](https://review.digital.arizona.edu/arizona-bootstrap/main/docs/5.0/components/background-wrappers/) component uses to a new `scss/img` directory, and added a one-line script (`css-copy-img`) to the normal build process defined in package.json, which copied the files into the `dist/css/img` directory, where the distributed CSS files could reference them. However PR https://github.com/az-digital/arizona-bootstrap/pull/233 removed the file copying operation, so there were then two independent copies of the SVG files, the ones actually used as part of the distributed assets in `dist/css/img`, and the supposedly canonical (but now disconnected) files under `scss/img`. Then PR https://github.com/az-digital/arizona-bootstrap/pull/1521 added the entire `dist/` directory subtree to `.gitignore`, with the intent that nothing there would be under direct version control, but would be a built artifact.

### To Reproduce

1. Check out a fresh copy of Arizona Bootstrap.
2. Delete everything within the `dist/` directory subtree.
3. Rebuild the CSS and JS files for distribution.
4. Observe that the SVG files do not appear in `dist/css/img`.

## Proposed resolution
As a stopgap, restore a file copying build step to populate `dist/css/img` with the SVG files; the copying step shouldn't be a simple duplicate of the one from https://github.com/az-digital/arizona-bootstrap/pull/157, but should allow for removed SVG files and spurious insertions to the directory by deleting its contents before copying the current files.

### Expected behavior
Current versions of the files needed for the background wrapper component should be included in the built artifacts within the `dist/' directory tree.

## Additional context
Upstream Bootstrap inlines the SVG needed to support some of the CSS files, but the code involved isn't large, unlike the Arizona Bootstrap SVG backgrounds, so that approach would bloat the code currently referenced anywhere that fully imports the Arizona Bootstrap CSS. However, currently the CDN-hosted distributions of Arizona Bootstrap require a separate transfer of each of the SVG files that the background wrapper component references. The backgrounds themselves might be slated for a major visual design change soon, so not much effort should be spent in providing them. Some of the SVG files currently present have been deprecated for a long time, and should be removed immediately.

mmunro-ltrr:
Actually updating the copied versions for now via Git, rather than letting them only be built from the originals.


mmunro-ltrr:
This is a possibly short-term improvement to part of the build process that over time had drifted away from having the SVG images in the `dist` directory subtree be derived from the canonical versions in `scss/img`. However there is a much more drastic long-term solution: moving the canonical versions of graphical elements out of Arizona Bootstrap entirely, so that they would be more accessible to software that might not otherwise use any form of Bootstrap (as already happens with the custom icons, though the repository for these is not necessarily the best home for the larger SVG background images and so on). Having the originals in a different place does not mean that they would be consumed through references to a CDN: Arizona Bootstrap builds could still make static copies of them, or even make embedded versions accessed as CSS variables.
Some arbitrary decisions are involved in the more limited current change from this PR:
- The `dist/css/img` directory gets destroyed and re-created before putting any files there.
- This has the side effect of removing some long-deprecated background images that no longer existed in `scss/img` (and means that future deletions in `scss/img` will propagate to `dist/css/img`).
- The SVG files are not simply copied, but processed with the [SVGO](https://svgo.dev/) optimizer; this seems a safe choice, since the icons repository already makes use of it, and if any problems emerge, there is scope for specifying a custom configuration in place of the SVGO defaults.

djcelaya (approved):
To test this, I deleted all contents from `dist/css/img/` locally and then rebuilt the project. Everything was regenerated into the `dist/css/img/` directory with no changes.

On another pass, I deleted just `trilines.svg` from the `scss/img/` directory and again rebuilt the project. I saw that the corresponding entry at `dist/css/img/trilines.svg` was deleted by the build process.

joeparsons (approved):

PR #2194, opened 2026-07-22
Closes #2174: Expose --az-navbar-fullscreen-height at :root for page-level layout rules

### LINKED ISSUE #2174: Expose `--az-navbar-fullscreen-height` at `:root` for page-level layout rules
## Problem/Motivation

The fullscreen navbar height token is not exposed at a scope that page-level layout rules can consume.

### Describe the bug

`--az-navbar-fullscreen-height` is currently only usable within styles scoped to the fullscreen navbar component. That makes it difficult to use for page-level spacing such as `body { padding-top: ... }`, because `body` sits outside that scope.

This leads to duplicated height values in downstream overrides and makes responsive/layout logic harder to maintain.

### Suggested fix

Expose a global height token at `:root`, for example:

```css
:root {
  --az-navbar-fullscreen-height: 69px;
}

@media (min-width: 996px) {
  :root {
    --az-navbar-fullscreen-height: 100px;
  }
}
```
### Expected behavior
`--az-navbar-fullscreen-height` should be available globally so page-level layout rules can consume it without duplicating hard-coded values or relying on component-local scope.

djcelaya:
## Changes

- Move `--az-navbar-fullscreen-height` to `:root`, with the existing responsive mobile override preserved.
Reason why: this makes the token available outside `.navbar-az-fullscreen`, which is required for page-level spacing rules such as `body` or main-content offsets.

- Update `.navbar-az-fullscreen-fixed-top-offset` to use `var(--az-navbar-fullscreen-height)` instead of hard-coded `100px` and `69px` values.
Reason why: this removes duplicated height values and keeps the helper aligned with the actual navbar token across breakpoints.

- Keep the remaining fullscreen navbar CSS variables scoped to `.navbar-az-fullscreen`.
Reason why: the issue only calls for exposing the height token globally, so this keeps the public token surface narrow and avoids widening component API unnecessarily.

- Refresh the fullscreen navbar docs text to clarify that `--az-navbar-fullscreen-height` is global, while the rest of the fullscreen navbar variables remain component-scoped.
Reason why: without the docs update, the current CSS variable section would describe the scope incorrectly.

## How to test
[Review site](https://review.digital.arizona.edu/arizona-bootstrap/issue/2174/)

1. Confirm no change to [AZ Navbar Fullscreen](https://review.digital.arizona.edu/arizona-bootstrap/issue/2174/docs/5.1/examples/navbar-az-fullscreen/). The [Campus Recreation](https://review.digital.arizona.edu/arizona-bootstrap/issue/2174/docs/5.1/examples/navbar-az-fullscreen/campus-rec/) page intentionally has a lengthy page to test the fixed navbar behavior.
2. Verify updated [docs subsection](https://review.digital.arizona.edu/arizona-bootstrap/issue/2174/docs/5.1/components/navbar/#css-variables) related to this root-level variable.

### DIFF
--- a/scss/custom/_navbar-fullscreen.scss
+++ b/scss/custom/_navbar-fullscreen.scss
@@ -1,7 +1,10 @@
-// Experimental custom az-fullscreen navbar
-.navbar-az-fullscreen {
+:root {
   // scss-docs-start navbar-az-fullscreen-css-vars
   --az-navbar-fullscreen-height: 100px;
+}
+
+// Experimental custom az-fullscreen navbar
+.navbar-az-fullscreen {
   --az-navbar-fullscreen-bg: var(--bs-blue);
   --az-navbar-fullscreen-bottom-border-color: var(--bs-red);
   --az-navbar-fullscreen-bottom-border-width: 3px;
@@ -95,11 +98,7 @@
 
 // Optional helper to offset content when using `.fixed-top` with fullscreen navbar.
 .navbar-az-fullscreen-fixed-top-offset {
-  padding-top: 100px;
-
-  @include media-breakpoint-down(lg) {
-    padding-top: 69px;
-  }
+  padding-top: var(--az-navbar-fullscreen-height);
 }
 
 .navbar-az-fullscreen-actions {
@@ -761,8 +760,11 @@
 
 /* Mobile Nav styling */
 @include media-breakpoint-down(lg) {
-  .navbar-az-fullscreen {
+  :root {
     --az-navbar-fullscreen-height: 69px;
+  }
+
+  .navbar-az-fullscreen {
     --az-navbar-fullscreen-brand-height: 45px;
     --az-navbar-fullscreen-toggler-icon-size: 24px;
     --bs-navbar-brand-padding-x: 12px;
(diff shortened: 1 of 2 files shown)


trackleft (dismissed):

scss/custom/_navbar-fullscreen.scss
```diff
@@ -1,7 +1,10 @@
-// Experimental custom az-fullscreen navbar
-.navbar-az-fullscreen {
+:root {
   // scss-docs-start navbar-az-fullscreen-css-vars
```
mmunro-ltrr: Move this comment (for including the code in the docs) to before the `:root {`?  (just a cosmetic thing).

mmunro-ltrr (dismissed):
There's one minor cosmetic thing about the code automatically included in the docs, but that's not important.

scss/custom/_navbar-fullscreen.scss
djcelaya: I had fixed this locally and apparently forgot to push it up. Thanks for the catch! Fixed with https://github.com/az-digital/arizona-bootstrap/pull/2194/changes/c21c1e054f9ef1853ef6ceaa5235334cd4cc4d07.

mmunro-ltrr (approved):

joeparsons (approved):

bberndt-uaz (approved):

trackleft (approved):

PR #2215, opened 2026-07-28
Closes #2203: AZ Navbar Fullscreen (mobile): Fix behavior for displaying initial menu page

### LINKED ISSUE #2203: AZ Navbar Fullscreen (mobile): Fix behavior for displaying initial menu page
### Describe the bug
In the initialization of the AZ Navbar Fullscreen JavaScript, if multiple menu links match the current page, the menu page for the last matching link (e.g., Resources by Audience) is displayed instead of the menu page for the first matching link (e.g., the main menu).

### To Reproduce

1. Create an example where multiple menu items from different menus have the same link as an item in the main menu.
2. Visit the page for the main menu.
3. Simulating a mobile device, observe that the main menu page is not displayed in the navigation.

## Proposed resolution
Correct the logic for displaying the active menu page.

bberndt-uaz:
## Changes

- Mobile: Updates JS functionality for showing the initial nav menu page:
   - Removes behavior which matched nav links with the browser URL. Instead, the initial nav menu page is determined solely based on nav links which have the `.active` class (which are added to the appropriate links in Quickstart).
   - Adds a check for an active link in the primary nav menu (which is displayed by default).
   - Reorders the secondary and tertiary nav link checks for an active link.
   - Removes the addition of the `.active` class to footer menu page links based on browser URL. Instead, the `.active` class is copied from the original nav link element (if it exists).
- Adds a new footer link to the Admissions example page for testing the initial mobile nav menu page behavior.

## Related issue
 - #2203

## How to test

1. Go to https://review.digital.arizona.edu/arizona-bootstrap/issue/2203/docs/5.1/examples/navbar-az-fullscreen/admissions/.
2. On mobile, confirm that the "Admissions" link in the primary menu has the sky color due to having the `.active` class.
3. Click on the Helpful Links footer and confirm that the "Duplicate Admissions link for testing" link also has the sky color due to having the `.active` class. This menu page is not the initial page since the primary page has priority.
4. Visit all the other example pages (you can use the links on the page content itself) and verify that the correct initial menu page is displayed for the mobile navigation.

_Note:_ An issue can arise if a link to the home page has NOT been added to the first level of the nav menu, but a link to the home page HAS been added to the second or third level of the nav menu or to a footer menu. Then, when a visitor loads the home page, the menu page containing that active link will be shown initially instead of the primary menu page. This seems like a content issue and so it is not addressed in this PR, but I'm open to any thoughts about this.

### DIFF
--- a/js/src/navbar-az-fullscreen-mobile-nav.js
+++ b/js/src/navbar-az-fullscreen-mobile-nav.js
@@ -51,11 +51,6 @@ class NavbarAzFullscreenMobileNav {
     this.initialMenuParentLabel = null
     this.initialMenuParentElementId = null
 
-    // Initialize window location variable
-    this.cleanWindowLocation = new URL(window.location.href)
-    this.cleanWindowLocation.search = ''
-    this.cleanWindowLocation.hash = ''
-
     this.init()
   }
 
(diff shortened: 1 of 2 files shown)


djcelaya (approved):
I pulled the HTML changes on this PR into a local copy of main to see [the error condition](https://github.com/user-attachments/assets/99ac8e66-d6a6-464c-9a6c-ffcebfeaef5f) and confirmed this does fix the issue.

I agree that we should not account for the case described in your note at this time–definitely a content problem IMO.

mmunro-ltrr (approved):
Checked on a phone.

trackleft (approved):

PR #2217, opened 2026-07-28
Closes #2216: modify Hugo's configuration file, allowing a version bump of hugo-bin to 0.161.0.

### LINKED ISSUE #2216: Version bump of hugo-bin from 0.149.2 to 0.161.0 breaks Dependabot pull requests.
## Problem/Motivation
All of the routine Dependabot dev dependency pull requests currently fail to make a review site because the static site builder fails at an initial checking step, long before there's even anything to deploy to AWS from GitHub.

### Describe the bug
Upstream Bootstrap v5 moved to Astro as its site building software, but Arizona Bootstrap continues to use its predecessor, Hugo. For a long time there were no changes to the hugo-bin npm package that provides this, but the software itself has continued to see rapid development: the package is just a wrapper for the Hugo binary, so that continued to evolve independently of the version numbers on the wrapper. Finally in the past few days the npm package has caught up (even its own release numbering has seen a big jump, from 0.149.2 to 0.161.0 with nothing inbetween). Unfortunately the new version of Hugo is incompatible with the current configuration files, so crashes at an early stage of building the documentation, which doubles as the target for the various automated checks.

### To Reproduce
Observe the GitHub actions logs for any recent Dependabot pull requests, such as
https://github.com/az-digital/arizona-bootstrap/actions/runs/30330999279/job/90185909790?pr=2213https://github.com/az-digital/arizona-bootstrap/actions/runs/30330999279/job/90185909790?pr=2213
```
ERROR error building site: assemble: failed to create page from pageMetaSource /docs/5.1: "/arizona-bootstrap-source/site/content/docs/5.1/_index.html:1:1": access denied: "text/html" is not whitelisted in policy "security.allowContent"; the current security configuration is:
 ...
```

## Proposed resolution
Either:
satisfy the requirements of the current Hugo version by modifying the configuration file,
or:
pin the hugo-bin package to the previously working version.

### Expected behavior
Dependabot pull requests should successfully update dev dependency package versions, pass code linting checks, and generate working review sites.

mmunro-ltrr:
Hugo introduced a new `allowContent` configuration setting in its v0.162.0
https://gohugo.io/configuration/security/#allowcontent
and development on the hugo-bin npm package started using this binary on 2026-05-26, The default setting is just `"! ^text/html$"`, documented as “By default, the HTML content format (media type text/html) is denied. Hugo emits HTML file content verbatim, which could allow arbitrary JavaScript execution.” But the Arizona Bootstrap documentation requires this, the security implications are minimal in practice, and a further explicit media type is needed to allow Markdown content.

### DIFF
--- a/hugo_template.yml
+++ b/hugo_template.yml
@@ -4,6 +4,8 @@ baseURL:                "{{az_site_host}}{{az_site_base_url}}"
 enableInlineShortcodes: true
 
 security:
+  allowContent:
+    - "^text/html$"
   enableInlineShortcodes: true
   funcs:
     getenv:
--- a/hugo_template.yml
+++ b/hugo_template.yml
@@ -6,6 +6,7 @@ enableInlineShortcodes: true
 security:
   allowContent:
     - "^text/html$"
+    - "^text/markdown$"
   enableInlineShortcodes: true
   funcs:
     getenv:
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "eslint-plugin-unicorn": "56.0.1",
     "find-unused-sass-variables": "^6.2.0",
     "globby": "^16.2.2",
-    "hugo-bin": "^0.149.2",
+    "hugo-bin": "^0.161.0",
     "jquery": "3.7.1",
     "js-yaml": "^5.2.2",
     "linkinator": "^7.6.1",


djcelaya (approved):
Thank you, @mmunro-ltrr, for this PR!! This clears CI but I did notice these deprecation warnings when I built this locally:

Image

Did you happen to see these in your build? Any concerns with these or should we ignore/address in a follow up?

joeparsons (approved):

PR #2010, opened 2026-05-01
Closes #1983: Use CSS variables wherever possible instead of hard-coded values

trackleft:
Review site: https://review.digital.arizona.edu/arizona-bootstrap/issue-1983/

- Convert color, border, and state styling in scss/custom/ and scss/override/ from Sass variables/property literals to Bootstrap CSS custom properties
- Refactor shared mixins in scss/mixins/ to consume CSS variables for buttons, callouts, backgrounds, and color contrast helpers
- Convert focus-ring Sass vars to CSS vars in _nav.scss, _carousel.scss, _dropdown.scss, and _photo-gallery.scss
- Convert border-radius Sass vars to CSS vars in _nav.scss and _button-group.scss
- Convert $breadcrumb-item-padding-x to var(--bs-breadcrumb-item-padding-x)
- Move button sizing, spacing, radius, box-shadow, and focus-box-shadow overrides into the .btn CSS variable layer in _buttons.scss
- Convert text-bg-* utilities to shared --bs-text-bg-color and --bs-text-bg-bg tokens
- Refactor navbar, badge, accordion, typography, utilities, and color-background overrides to use token-driven CSS vars instead of Sass-only values
- Remove now-unused Sass variable definitions/imports made obsolete by the CSS variable migration
- Refactor docs site SCSS in site/assets/scss/ to use docs-level CSS custom properties for brand, link, anchor, button, navbar, skippy, TOC, and color token styling
- Add docs-scoped variables in site/assets/scss/_variables.scss, including --bd-purple-light and related docs token definitions
- Update the Sass customization docs to reflect the new CSS variable approach
- Regenerate compiled dist CSS assets

Validation:
- Ran Playwright visual regression across all captured docs/example pages; 189/189 tests passed
- Ran css-lint successfully after the refactor

Closes #1983

PR #2225, opened 2026-07-30
Manual visual regression testing with Playwright

### LINKED ISSUE #2226: Manual visual regression testing with Playwright
I would like a visual regression tool to use with the development of 5.2.

A while back, Chris created #2010 and claimed his Playwright VR tests passed. We do not have VR testing in this repo that I am aware of so I asked about his process for doing so and received the following:

- [playwright.config.ts](https://github.com/user-attachments/files/30596597/playwright.config.ts)
- [tests.zip](https://github.com/user-attachments/files/30596600/tests.zip)

My attempts at recreating this workflow are in #2225.

djcelaya:
Extension of some work Chris did earlier this year to implement local visual regression testing with Playwright. The immediate goal of this PR is to assist in the review of #2010. My hope is that this tool can be committed to the repository for long-term use with other PRs. An evolution of this project would be including it as a non-blocking PR check.

This is a WIP.

---

## What this adds

**Visual regression (VR) testing** built on [Playwright](https://playwright.dev/), running **entirely inside the existing Docker container**. It exports two branches, builds the docs site and screenshots every page for each — in Linux — then compares them to surface unintended visual changes. Because it runs in the container, results are independent of the host OS/toolchain and baselines are reproducible.

- `Dockerfile` — the shared image now also installs Chromium (`npx playwright install --with-deps chromium`) and a `vr-run` entrypoint, so VR uses the same image as the review site.
- `scripts/vr-test.sh --base <branch>` — the single entry point: exports the base branch and current `HEAD` with `git archive` (no checkout), builds/screenshots both in the container, and compares.
- `scripts/vr-run.sh` — the in-container entrypoint (`/usr/local/bin/vr-run`): links the image's `node_modules`, builds `dist` + `docs-build`, runs Playwright.
- `scripts/build-container-image.sh` — image build/ensure logic extracted from `build-container.sh` so the review site and VR share one image.
- `playwright.config.ts` — Chromium-only; `fullyParallel: true`; serves each build on VR port **9002** (override `VR_PORT`) inside its own container run; snapshots directed to a shared dir via `VR_SNAPSHOT_DIR`.
- `tests/` — the spec (`visual-regression.spec.ts`), page discovery (`utils/discover-pages.ts`), and a README.
- npm: `test:visual` → `scripts/vr-test.sh` (pass `--base <branch>`).

## Requirements

- **Docker** (Docker Desktop on macOS/Windows, or Docker Engine on Linux).
- Nothing else on the host — no Node, no browser install. The container provides Node, the toolchain, Hugo, and Chromium. The image builds automatically on first use (and rebuilds when the `Dockerfile`, `scripts/`, or lockfile change) and is shared with `scripts/build-container.sh`.

## How to use it

Compare the current branch (`HEAD`) against a base branch:

```bash
scripts/vr-test.sh --base 5.2.x    # compare a 5.2 feature against 5.2.x
scripts/vr-test.sh --base main     # or any branch/ref

# equivalently:
npm run test:visual -- --base 5.2.x
```

It exports both branches, builds and screenshots each in the container, and reports differences (non-zero exit on any diff). It's safe to run in a **second terminal alongside a running `scripts/build-container.sh`** review site — it never switches your working tree.

View a failing run's report on the host (needs Node ≥ 20) or open it directly:

```bash
npx playwright show-report      # or: open playwright-report/index.html
```

On failure, `-actual` / `-expected` / `-diff` images are written to `test-results/` and an HTML report to `playwright-report/` (both gitignored).

> VR compares **committed** state (`HEAD`); commit local changes first to include them.

## Intended workflow

1. Check out the feature branch you want to review (e.g. #2010).
2. Commit any local changes (VR compares committed state).
3. Run `scripts/vr-test.sh --base 5.2.x` (you can leave the review-site container running — VR is parallel-safe).
4. Inspect any failures in the HTML report — real visual changes vs. expected design changes.
5. Intentional changes simply become the new expected result on the next run's base seed; nothing is committed (baselines are ephemeral).

## How to test

You can validate the tool end to end without a real regression:

**1. Self-comparison — should report *no* differences.** Comparing a branch against itself exercises the full export → build → seed → compare flow; identical source on both sides means zero diffs:

```bash
# while on manual-vr-test:
scripts/vr-test.sh --base manual-vr-test
# → 191 passed, "no differences from 'manual-vr-test'"
```

**2. Branch off and introduce a change — should *surface* the diff.** Branch from `manual-vr-test`, make a visible change to a Bootstrap source file, commit it, and compare back against `manual-vr-test`:

```bash
git switch -c my-vr-check manual-vr-test
# edit a file under scss/ or js/src/ so a docs page renders differently
git add -A && git commit -m "test: visual change"
scripts/vr-test.sh --base manual-vr-test
# → the affected pages fail (non-zero exit)
open playwright-report/index.html      # or: npx playwright show-report (needs Node ≥ 20)
```

This confirms both directions: no false positives on identical source, and real changes flagged and scoped to exactly the pages they affect.

## Design decisions & rationale

- **Runs in the container, not on the host.** The review-site workflow (`build-container.sh`) installs Linux `node_modules` onto the host, which breaks host-native builds on macOS/arm64 (wrong-platform `rollup`/`hugo` binaries). Running VR in the container removes the platform mismatch and matches the project's "everything builds in the container" convention.
- **One shared image, many entrypoint scripts.** Chromium and `vr-run` are added to the existing image (alongside `serve-review-site`, `lint`, etc.) rather than a separate image, following the established pattern. Image-ensure logic is extracted to `build-container-image.sh` so the review site and VR share it (DRY).
- **`git archive`, not checkout.** Both branches are exported into throwaway temp dirs, so the working tree is never switched. VR is therefore **safe to run in a second terminal** while the review-site container live-watches the tree — no branch-switching conflicts, and no port collisions (VR's Hugo runs on 9002 inside its own container).
- **Baselines are ephemeral and reproducible.** Nothing is committed: each run seeds fresh baselines from the base branch into a temp dir, compares, and discards them. Captured in Linux, they're reproducible (`-chromium-linux`), which makes CI practical later.
- **Each branch builds with its own `package.json`.** `vr-run` links the image's `node_modules` without overwriting the exported tree's `package.json`/lockfile, so version strings render per-branch and comparisons stay accurate.
- **Parallelized.** `fullyParallel: true` spreads the ~191 screenshots across workers (≈4× faster locally: ~2.6m vs ~11.5m); CI stays pinned to a single worker for determinism.

## Known caveats

- **Sub-pixel flakiness.** The comparison threshold is `maxDiffPixels: 100`. Anti-aliasing/font-hinting nondeterminism can occasionally tip a page just over the threshold — a false positive, not a regression. If noisy, loosen the tolerance (`maxDiffPixelRatio` or a per-pixel `threshold`).
- **Assumes the base branch contains the tooling.** Seeding from the base branch requires the VR tooling to exist there — true once this PR is merged into the base line (e.g. `5.2.x`).
- **Report viewer needs host Node ≥ 20.** `npx playwright show-report` runs on the host; use Node ≥ 20 or `open playwright-report/index.html`.
- **Ownership / perf.** On Linux hosts, container-written artifacts may be owned by uid 1000 (remapped automatically on macOS Docker Desktop). On macOS, screenshotting every page over a bind mount is slower than native.

## Validation

- Self-comparison (a branch against itself): **191 passed, zero diffs** in ~2.6m — confirms the export/build/seed/compare flow and that identical source produces no differences.
- Real feature branch vs. `manual-vr-test`: **182 passed, 9 failed**, with failures correctly scoped to exactly the pages the change affected (the `navbar-az-fullscreen` examples) and a non-zero exit.

## Future work

- **CI integration** — now practical because the Linux baselines are reproducible; a natural next step is a non-blocking PR check.
- Additional browser/viewport projects.


djcelaya:
This unfortunately is looking like it's going to take a lot longer than I had hoped to get this to MVP. Closing for now but might pursue again in the future.

PR #2112, opened 2026-06-17
Closes #1975: Add new colors needed for arizona.edu 2026 redesign

djcelaya:
See #1975.

Updates 5 colors derived from [the arizona.edu redesign project](https://github.com/uaz-web/az-storybook/blob/main/packages/components-sdc/css/color.css). Also updates Color contrast matrix to maintain a minimum of WCAG AA color contrast between text color and background color.

## How to test
[Review site](https://review.digital.arizona.edu/arizona-bootstrap/issue/1975/)
1. Compare new [Color](https://review.digital.arizona.edu/arizona-bootstrap/issue/1975/docs/5.1/customize/color/#all-colors) page with old [Color](https://review.digital.arizona.edu/arizona-bootstrap/main/docs/5.1/customize/color/#all-colors) page.
2. Compare new [Color Contrast](https://review.digital.arizona.edu/arizona-bootstrap/issue/1975/docs/5.1/getting-started/color-contrast/) page with old [Color Contrast](https://review.digital.arizona.edu/arizona-bootstrap/main/docs/5.1/getting-started/color-contrast/) page.
3. Compare new [Background](https://review.digital.arizona.edu/arizona-bootstrap/issue/1975/docs/5.1/utilities/background/) page with old [Background](https://review.digital.arizona.edu/arizona-bootstrap/main/docs/5.1/utilities/background/) page.
4. Check various other pages using or displaying these updated colors.

## Color contrast updates

The Color Contrast docs page is driven by a hand-maintained data file ([`site/data/contrast.yml`](https://github.com/az-digital/arizona-bootstrap/blob/issue/1975/site/data/contrast.yml)) rather than computed at build time — the Hugo template simply flags any combination not listed there as inaccessible. Because those accessible combinations were curated against the previous hex values, the table had to be re-evaluated for the redesigned palette. Re-running the WCAG AA math (≥ 4.5:1 for normal text) across every pairing that involves a changed color surfaced the flips below.

### oasis (`#378dbd` → `#106ab1`)

| Foreground / background | Old | New |
| --- | --- | --- |
| oasis text on white | 3.68:1 ❌ | 5.65:1 ✅ |
| white text on oasis | 3.68:1 ❌ | 5.65:1 ✅ |
| midnight text on oasis | 4.53:1 ✅ | 3.27:1 ❌ |
| warm-gray text on oasis | 3.17:1 ❌ | 4.92:1 ✅ |
| cool-gray text on oasis | 2.99:1 ❌ | 4.85:1 ✅ |
[...]

PR #2256, opened 2026-08-12
Closes #2250, #2244: Update breadcrumb styling from arizona.edu redesign

### LINKED ISSUE #2244: Breadcrumb icons are not accessible
The icon is not accessible because a pseudo-element is not a normal HTML element and cannot reliably have `aria-hidden="true"` applied to it. We should replace the icon-font Unicode character that is displaying this icon in CSS content with a background or mask that provides the same icon.

### LINKED ISSUE #2250: Update breadcrumb styling for arizona.edu 2026 redesign
## Motivation
The breadcrumbs on the arizona.edu 2026 redesign site have received some visual updates. We would like to make this new style available to campus as the default breadcrumb style.

We would like to keep breadcrumb styling consistent across our sites.

Replace existing breadcrumb styling so that it matches arizona.edu.

Provide a breadcrumb styling variant.

## Additional context
- [Here is the breadcrumb SDC](https://github.com/uaz-web/az-storybook/tree/main/packages/components-sdc/components/breadcrumb) used on the new arizona.edu site.
- [Breadcrumb SDC CSS](https://github.com/uaz-web/az-storybook/blob/main/packages/components-sdc/components/breadcrumb/breadcrumb.css)
- [AZBS Custom `.breadcrumb-item-home` class](https://github.com/az-digital/arizona-bootstrap/blob/main/scss/custom/_breadcrumb.scss)
- [AZBS Breadcrumb overrides](https://github.com/az-digital/arizona-bootstrap/blob/main/scss/override/_breadcrumb.scss)
- We might consider addressing #2244 with this request as well.

kevdevlu:
See #2250. Also closes #2244.

Replaces the existing breadcrumb styling to match the [arizona.edu 2026 redesign's breadcrumb SDC](https://github.com/uaz-web/az-storybook/tree/main/packages/components-sdc/components/breadcrumb), following the same in-place replace pattern used for the redesign's color updates in #2112 (as opposed to adding an opt-in variant).

## Changes

1. Match arizona.edu redesign breadcrumb styling. See [adobe xd mockup](https://xd.adobe.com/view/f28d87c5-418d-4c27-b7cc-627e99ca2c14-a7de/screen/c42e68c5-7f13-4e2d-8c47-bb7669dd7b26/specs/?hints=off).

2. Fixed 2 underline issues that you can see on arizona.edu right now:
    Compare a regular breadcrumb item
    image
    to the breadcrumb home underline:
    image
    The breadcrumb home underline is noticeably closer to the text.

    Also if you look at the `Home` underline in the [mockup](https://xd.adobe.com/view/f28d87c5-418d-4c27-b7cc-627e99ca2c14-a7de/screen/c42e68c5-7f13-4e2d-8c47-bb7669dd7b26/specs/?hints=off), the underline doesn't extend pass the home icon:
    image
    **whereas in https://www.arizona.edu/about/rankings it does extend past the home icon:**
    image
So this PR tried to address both of these

3. The breadcrumb `home` icon needs to be FILL: 0 instead of FILL: 1. (Although doing point 4. below – the accessibility issue – meant that i had to switch it to a SVG, so this might not matter for us, but i kept the change just on principle of we're now allowing outlined icons instead of just filled icons)
    In current `main`, arizona bootstrap has:
    [`opsz,wght,FILL,GRAD@24,400,1,0`](https://github.com/az-digital/arizona-bootstrap/blob/0f32c8fd279ed117127cadd6cfb64d1ff66489fa/site/layouts/partials/stylesheet.html#L1)

    In current storybook `main`, we have:
    [`opsz,wght,FILL,GRAD@24,100..700,0,0`](https://github.com/uaz-web/az-storybook/blob/ca743db1f69d51b8bc6905ce9c1d6fc193a8ff90/packages/storybook/.storybook/preview-head.html#L24)

    So this PR combines these:
    [`opsz,wght,FILL,GRAD@24,100..700,0..1,0`](https://github.com/az-digital/arizona-bootstrap/blob/4e1f5d2e2da63afaade0ca8f1fb597e45f1d9a32/site/layouts/partials/stylesheet.html#L1)

4. Closes #2244: the home icon was rendered via a Material Symbols codepoint in CSS `content`, which can't take `aria-hidden` since a pseudo-element isn't a real DOM node — so screen readers could read the raw glyph as text. Replaced it with an inline background SVG instead (the same shape Google serves for Rounded/home/wght300/24px), which isn't announced at all. 

## How to test

[Review site](https://review.digital.arizona.edu/arizona-bootstrap/issue/2250/) → [Breadcrumb component doc page](https://review.digital.arizona.edu/arizona-bootstrap/issue/2250/docs/5.1/components/breadcrumb/)

See if the Breadcrumb font weights, hover underlines, icons match
- [Adobe XD mockup](https://xd.adobe.com/view/f28d87c5-418d-4c27-b7cc-627e99ca2c14-a7de/screen/09eee508-4a4e-4846-bdfe-36da2eee644f/?hints=off),
- [arizona.edu pages](https://www.arizona.edu/about/rankings), and
- [storybook](https://github.com/uaz-web/az-storybook/tree/main/packages/components-sdc/components/breadcrumb)

For #2244 specifically: inspect the home icon in devtools and confirm it's a `background-image`, not `content`, on `.breadcrumb-item-home::before` — and that no glyph text shows up in the accessibility tree:
Before:
image
After:
image


scss/override/_breadcrumb.scss
```diff
@@ -1,17 +1,50 @@
...
     }
   }
 }
 
+.breadcrumb-item {
+  line-height: 1;
+}
+
+// Extra bottom padding matches this <li>'s box height to `.breadcrumb a`'s.
+.breadcrumb-item-home {
```
djcelaya: Can we move all the definitions for `.breadcrumb-item-home` to `custom/_breadcrumb.scss`? This class does not exist upstream so to help with organization, we try to keep new things we add in the `custom` directory. If the class exists upstream in BS5 and we're changing it (override), these would go in `override`.

site/content/docs/5.1/icons.md
djcelaya, line 1: Since the icon is now a fully-contained SVG, do we even need the Material Symbols requirement anymore?

site/layouts/partials/stylesheet.html
djcelaya, line 1: I was concerned this might effect other icons in the project but I haven't seen any regressions yet.

djcelaya (changes requested):
This is looking great. Thanks, @kevdevlu!

djcelaya:
Feedback from Tom: Reduce x-padding of the `.breacrumb` so that its content more closely align with the left edge of the text in the nav-item above. This may require some eye-balling. It's maybe about 12px (down from 16). This is extra tricky because the icon appears to have some baked in spacing that may be contributing to the visual misalignment. 

Screenshot 2026-08-13 at 4 34 04 PM

scss/override/_breadcrumb.scss
kevdevlu: moved!

site/content/docs/5.1/icons.md
kevdevlu: I'll take it out and revert to be same as `main`

kevdevlu:
> Feedback from Tom: Reduce x-padding of the `.breacrumb` so that its content more closely align with the left edge of the text in the nav-item above. This may require some eye-balling. It's maybe about 12px (down from 16). This is extra tricky because the icon appears to have some baked in spacing that may be contributing to the visual misalignment.
> 
> Screenshot 2026-08-13 at 4 34 04 PM

I made the change and applied it to this multidev: https://placeholder-azs-portuguese-flagship.pantheonsite.io/test-section/test-page. 
image

djcelaya:
This is approved by Marketing on Aug 14.

djcelaya (approved):

mmunro-ltrr (approved):

PR #2258, opened 2026-08-12
Suggested changes to fix badge background.

### LINKED ISSUE #1983: Use CSS variables wherever possible instead of hard-coded values.
## Motivation
It would be great if we could override arizona-bootstrap by changing CSS variables instead of having to override a specific class.

For example if arizona-bootstrap does something like:
```
.btn-red {
background-color: #ab0520;
}

.bg-red {
  background-color: #ab0520;
}
```

If we want to override the background-color we need to do:

```
.btn-red {
   background-color: #ab0521;
}

.bg-red {
  background-color: #ab0521;
}
```

However if we used css variables, like:
```
.btn-red {
  background-color: var(--bs-red)
}
```
We could override a button like :

```
.btn-red {
  --bs-red: #ab0521;
}
```

Or we could override all like

```
:root {
  --bs-red: #ab0521;
}
```
Or 
```
html {
  --bs-red: #ab0521;
}
```

I want to change the red everywhere.

## Proposed Resolution

Update the boostrap css to use this kind of styles
```
.btn-red {
    background-color: var(--bs-red)
}
```

Update the boostrap css to use this kind of styles
```
.btn-red {
    background-color: var(--bs-red)
}
```

N/A

trackleft:
https://review.digital.arizona.edu/arizona-bootstrap/refactor-text-bg-color/docs/5.1/components/badge/


djcelaya:
Thanks, @trackleft! Unfortunately, I'm still seeing a regression with the Light badge link on hover. I think this PR will fix that particular problem: https://github.com/az-digital/arizona-bootstrap/pull/2264

Additionally, I'm seeing quite a few regressions related (I think) to the contrast changes on this PR. Should I try and fix those? I'm unsure as to why there are so many color contrast changes here TBH.
Screenshot 2026-08-18 at 3 21 52 PM

trackleft:
Addressed in #2010

PR #2265, opened 2026-08-18
Closes #2159: AZ Navbar Fullscreen (mobile): Add fade effect at the top of scrolling menu items

### LINKED ISSUE #2159: AZ Navbar Fullscreen (mobile): Add fade effect at the top of scrolling menu items
This pull request updated the mobile display of the AZ Navbar Fullscreen to have a fade effect above the top footer, just like on desktop:
 - #2153

That change allowed the bottom of the list of nav menu items to fade out, indicating that further menu items are available if the user scrolls. For this issue, an equivalent fade effect should also be added for the top of the list of nav menu items. This behavior requires different functionality: the fade effect should only appear once the user has scrolled down the list. From initial investigation, implementing this top fade effect is not possible with pure CSS, so some additional JavaScript is required.

## Conditions of satisfaction
- [x] No fade effect should be visible when the menu items are scrolled to the top (such as on initial load).
- [x] A fade effect should be visible when the menu items have been scrolled down.
- [x] The fade effect should match the appearance of the one above the footer.

bberndt-uaz:
## Changes

1. Mobile: Adds a `setupMobileMenuObserver()` function to `navbar-az-fullscreen-mobile-nav.js` which initializes an intersection observer to determine if the user has scrolled the mobile nav menu. If so, a "scroll shadow" or fade effect is added to the top of the nav menu.
2. Mobile: Adds a new `navbar-az-fullscreen-nav-mobile-menu-container` div as a container for the `.navbar-az-fullscreen-modal-menu-nav-col` divs on mobile, to enable the intersection observer to work properly. This div element is added both to the JavaScript and to the example HTML.
3. Mobile: Adds styling for the new scroll shadow at the top of the nav menu via the `.show-shadow` class.
4. Moves the fade effect for the bottom of the menu nav from the top footer element to the bottom of the `.modal-body` element, so it will be displayed even if the top footer is missing.
5. Mobile: Moves the top separator line for the mobile menu nav from the menu nav element to become a top border of the new `.navbar-az-fullscreen-nav-mobile-menu-container` element.
6. Mobile: Removes the `padding-right: 0` override for `.navbar-az-fullscreen-modal-menu` on mobile and adds a negative margin to `.navbar-az-fullscreen-modal-menu-nav-col` to ensure the scroll bar remains flush with the right edge of the screen.
7. Mobile: As a nice side effect of the above change, the call-to-action items are now correctly centered horizontally!
8. Adds a new `--az-navbar-fullscreen-mobile-padding-xy` CSS variable to replace hard-coded 20px values.
9. Removes `flex: 0 0 auto` included as part of the style attribute which is added to menu nav elements by `navbar-az-fullscreen.js`. This CSS property is either (a) already present for the element from a Bootstrap column class or (b) doesn't appear to be necessary. I added it to `.navbar-az-fullscreen-modal-menu-nav-col-tertiary` for desktop (see line 591) but I think it may not be necessary.
10. Related to the above, the example HTML has been updated to remove mobile-only Bootstrap column classes. The exception is the `.col` class for both desktop and mobile that goes with `.navbar-az-fullscreen-modal-menu-nav-col-tertiary`: the flex property mentioned in number 8 replaces this. 

## Related issues
 - #2159

## How to test

1. Visit https://review.digital.arizona.edu/arizona-bootstrap/issue/2159/docs/5.1/examples/navbar-az-fullscreen/.
2. On desktop, compare the changes with the [main branch](https://review.digital.arizona.edu/arizona-bootstrap/main/docs/5.1/examples/navbar-az-fullscreen/) and confirm that there are no visual changes. I recommend opening two tabs and comparing them as you navigate the menu.
3. Simulating a mobile device, confirm there are no visual changes if you do not scroll on any nav menu page EXCEPT FOR the corrected horizontal centering of the call-to-action items.
4. Confirm that the top scroll shadow / fade effect has been added as intended:
   - Confirm that the top scroll shadow does not appear unless you scroll the list of menu items.
   - Confirm that the top scroll shadow appears for primary, secondary, tertiary, and footer menu pages. (You may need to go into DevTools and duplicate a menu item to make the list long enough to scroll.)
   - Confirm that the top scroll shadow is the same as the bottom (except inverted).


djcelaya (commented):
I think I found a behavior change on desktop. Here's the starting scenario:

Image
I have constrained the viewport vertically so the first and second columns need scroll. I have scrolled the middle column to the bottom and intend to open the Health & Wellness submenu. When I hit the nav toggle, the middle column scrolls itself to the top as the tertiary column is shown and my cursor is now hovering a different nav-link. This does not happen on main as far as I can tell.

Image

djcelaya (commented):
Mobile appears to be working just fine. I followed the testing instructions and confirmed the fade effect is added correctly for all the various menu columns. CTA center looks good. Fade effect appears correct for top and bottom.

bberndt-uaz:
> I think I found a behavior change on desktop. Here's the starting scenario:
> 
>  I have constrained the viewport vertically so the first and second columns need scroll. I have scrolled the middle column to the bottom and intend to open the Health & Wellness submenu. When I hit the nav toggle, the middle column scrolls itself to the top as the tertiary column is shown and my cursor is now hovering a different nav-link. This does not happen on main as far as I can tell.

Thank you! I see the issue too. I'll take a look.

bberndt-uaz:
@djcelaya I believe my latest commit has fixed the issue of the secondary column auto scrolling to the top. The column seems to require `flex: 1 1 auto` and it appears that the `flex: 0 0 auto` added by the JavaScript was not having an effect?

djcelaya:
This is approved by Marketing on Aug 20.

djcelaya:
> @djcelaya I believe my latest commit has fixed the issue of the secondary column auto scrolling to the top. The column seems to require `flex: 1 1 auto` and it appears that the `flex: 0 0 auto` added by the JavaScript was not having an effect?

I suspect it was preventing this from happening in some way 😬 :
Screenshot 2026-08-20 at 2 36 02 PM
This is regular desktop with no scroll needed.

I can verify that the auto scroll issue is fixed though.

djcelaya:
See #2272 for a possible fix for the regression I posted above.

djcelaya:
Another possible issue. When I'm in mobile (on first page load) and open up the modal menu, if the primary nav is shown and needs scroll, the top effect is not applied. However, if I navigate to a secondary level item (still on the same page load) and then back to primary and scroll, the top fade appears.

So start condition is I'm on the Home page and open the mobile menu. Scroll is needed and I see this:
Screenshot 2026-08-20 at 3 23 21 PM
I click/tap into the Admissions submenu and then immediately back to the primary level and see:
Screenshot 2026-08-20 at 3 28 36 PM

If I'm looking at a secondary or tertiary level page Admissions > Undergraduate or Admissions > Undergraduate > Tuition & Aid, and open the modal menu for the first time, the top fade correctly appears if scrolled.

bberndt-uaz:
@djcelaya Thank you very much for your detailed review with the helpful screenshots! I neglected to add the "sentinel" element for the top scroll shadow to the example pages HTML. The initial primary nav should have the top fade now.

Thanks also creating that other PR with the fix for the other regression: I'm happy to merge it anytime. Did you want to accept that Copilot suggestion for the comment?

djcelaya:
> @djcelaya Thank you very much for your detailed review with the helpful screenshots! I neglected to add the "sentinel" element for the top scroll shadow to the example pages HTML. The initial primary nav should have the top fade now.
> 
> Thanks also creating that other PR with the fix for the other regression: I'm happy to merge it anytime. Did you want to accept that Copilot suggestion for the comment?

No problem! Yes, I just applied the comment update. Merge away!

djcelaya (dismissed):

djcelaya (approved):
Looks good! Also checked on a real phone this time.

trackleft (approved):

mmunro-ltrr (approved):
Checked on a real phone.

Issue #1764, opened 2025-10-06
Ensure all supported classes are included in the docs


bberndt-uaz:
Our documentation site for Arizona Bootstrap 5 does not include all supported classes for certain utilities and components. So far, the classes we have identified as missing from the documentation are all color variant classes, but other classes might be missing as well.

The goal is for the documentation to transparently present exactly what classes are available in Arizona Bootstrap. If certain classes are discouraged and should be used sparingly or only in special circumstances, notes should be added for these guidelines. If an undocumented class is included in Arizona Bootstrap which should not be, we should remove it.

Examples of classes missing on certain pages:

 - [Borders](https://review.digital.arizona.edu/arizona-bootstrap/main/docs/5.0/utilities/borders/): `.border-black`
 - [Buttons](https://review.digital.arizona.edu/arizona-bootstrap/main/docs/5.0/components/buttons/): `.btn-sky`, `.btn-primary`, `.btn-secondary`, `.btn-success`, `.btn-dark`
 - [Colors](https://review.digital.arizona.edu/arizona-bootstrap/main/docs/5.0/utilities/colors/): `.text-primary`, `.text-secondary`, `.text-success`, `.text-dark`

**Note:** Classes for badges can be handled with this existing issue:
 - #1738 

## Conditions of satisfaction
- [ ] Add missing supported classes to the documentation
- [ ] Add notes to the docs about any classes for which usage is recommended or discouraged
- [ ] Remove any undocumented classes from Arizona Bootstrap which should not be supported
- [ ] Add "Custom Arizona Bootstrap" badges where appropriate

PR #2281, opened 2026-08-26
Closes #1956: Update buttons with 2026 brand changes

### LINKED ISSUE #1956: Update buttons with 2026 brand changes
## Motivation
The flagship arizona.edu site is undergoing a redesign as indicated in these Adobe XD mockups:

Components: https://xd.adobe.com/view/5e5859af-459f-438c-83e0-963b6791dd91-0b0b/grid
Desktop Nav: https://xd.adobe.com/view/72056437-64ab-4b34-9c61-ade0cf25b6c6-2756/grid
Mobile Nav: https://xd.adobe.com/view/df468591-06c4-4834-ab38-7920897e9c57-8a15/grid

The [Buttons screen](https://xd.adobe.com/view/5e5859af-459f-438c-83e0-963b6791dd91-0b0b/screen/70b91355-49a6-4725-952e-7c02aecc9edd/) included in the Components link above defines new styles for buttons.

Buttons in storybook: https://uaz-web.github.io/az-storybook/?path=/docs/components-button--buttons

The new design makes the following changes to the default Arizona Bootstrap buttons:

1. Focus color added (e.g., chili for .btn-red). Current buttons do not change color with focus.
2. Focus outline added (`outline: 2px solid var(--bs-bloom); outline-offset: 2px;`). 
3. Padding top and bottom changed from 8px to 12px. (Need to consider line height and button borders with this change.)
4. Padding left and right changed from 20px to 30px. (Need to consider button borders with this change.)
5. Border radius changed from 48px to 24px.
6. Font changed from Proxima Nova to Proxima Nova Condensed.
7. Font weight changed from 500 to 700.
8. Font size changed from 16px to 18px.
9. Line height changed from 24px to 22px.
10. Letter spacing property added (.14px).

Further changes are made to Arizona Bootstrap `.btn-lg` buttons:

1. Padding top and bottom changed from 8px to 14px.
2. Padding left and right changed from 16px to 36px.
3. Border radius changed from 48px to 30px.
4. Font size changed from 20px to 22px.
5. Line height changed from 30px to 27px.
6. Letter spacing property added (.18px).

In addition, a new style of button is also present in the Desktop Nav mockups: note the "Become a Wildcat" button on many of the pages. This new button style has the following custom styling (asterisks indicate differences from the default buttons above):

1. *padding-top and -bottom: 10px
2. padding-left and -right: 30px
3. *border-radius: 20px
4. font-family: Proxima Nova Condensed
5. font-size: 18px
6. *font-weight: 600 
7. line-height: 22px

These values should be verified and we should confirm whether any of them could be slightly different (e.g., 32px instead of 30px, to potentially use existing classes or Sass variables). One adjustment which could produce the same end result is making the line height 1.25rem and setting padding-top to 10px (to account for Proxima Nova Condensed descender space).

### Questions

1. Which of the new button styles (if any) need to be implemented as brand changes for all buttons in Arizona Bootstrap? For example, should all Arizona Bootstrap buttons be updated with a focus color and/or focus outline?
   - In general, these button updates should be implemented as brand changes.
2. Does the button styling in the Navigation mockups need to be considered as another variant, one that is different from the buttons included in the Component designs?
3. Is the "Focus State" for buttons intended to apply on `:focus` (no existing Arizona Bootstrap styling) or `:focus-visible` (existing focus ring in Arizona Bootstrap)?
   - Focus states should probably be the same for both click state and keyboard navigation.
4. Is it important for the new buttons to be changed to have a total height of 46px rather than the Arizona Bootstrap default of 44px?
5. The new button styles do not include borders. Do the default button borders from Bootstrap need to be removed?
6. Is the new border radius acceptable for buttons with larger text or multiple lines of text?
   - We can keep the Arizona Bootstrap default for now and review it further in the future.
7. The button hover state transition indicates "dissolve": this appears to be an Adobe XD transition type. Is the Bootstrap default `ease-in-out` CSS [transition timing function](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/transition-timing-function) acceptable?
8. The Featured Event Component screen includes an Outlined Button Sky. (In fact, Arizona Bootstrap already includes a .btn-outline-info which uses Sky, but the text color is incorrect.) Are there additional Outlined Buttons colors that should be supported?
   - Colors represented on the Buttons screen are the only ones that should be available. Outlined Button Sky is in fact represented on that screen.
   - Restrict options for button colors based on background.  - Blue button could also be included on a white background.

Although it would be possible to simply create a new button variant that contains all the styles listed above, it would be preferable to add additional classes and SCSS variables to support some of the CSS changes above. For example, we could add new classes for font-family, font-size, line-height, and border-radius. For font-weight, we could use the existing `fw-*` classes.

Any button classes we add should follow the precedents for existing button styles in Arizona Bootstrap. For example, the new padding styles could be included as part of the styling for a new `.btn-az-lg` button variant. On the other hand, buttons should allow for any font to be used, so a separate class should be used to set the font (if it will be different from the default font family for buttons).

bberndt-uaz:
## Changes

_Note:_ These changes are intended to update Arizona Bootstrap buttons so that they closely (but not *exactly*) match the appearance of the [button component in az-storybook](https://uaz-web.github.io/az-storybook/?path=/docs/components-button--buttons) (and on the new arizona.edu site). We'll need a follow-up PR for the [buttons in the az-storybook repo](https://github.com/uaz-web/az-storybook/tree/main/packages/components-sdc/components/button) to remove the Arizona Bootstrap v5.1.x overrides. See the [original Adobe XD designs for the button component](https://xd.adobe.com/view/5e5859af-459f-438c-83e0-963b6791dd91-0b0b/screen/70b91355-49a6-4725-952e-7c02aecc9edd/) for further consideration.

1. Updates styling of buttons in accordance with the [button component in az-storybook](https://uaz-web.github.io/az-storybook/?path=/docs/components-button--buttons):
   - Sets the font-family to Proxima Nova Condensed.
   - Updates the padding values for standard, large, and small button sizes.
   - Rather than using the new `text-box` CSS property (see the [caniuse page](https://caniuse.com/css-text-box-trim)), button top and bottom padding are adjusted to vertically center the button text. As a result, button heights are whole integer values which match the Adobe XD designs.
   - Adds line-height and letter-spacing properties.
   - Overrides existing box-shadow `:focus-visible` styling in favor of an outline around the button.
   - Maintains the existing 2px border to avoid sizing issues with the outline variants.
2. Adds supported button color variants to the Buttons docs page which were previously missing. (See #1764 for additional cases.)
3. Adds new Sky, White (Text Red), White (Text Blue), and Outlined Sky button variants.
4. Refactors solid and outline button SCSS, including the addition of a new az-button-outline-variant() mixin.
5. Reorders some content on the Buttons docs page.

## Questions

1. Is it correct that if we are going to remove any button variants, we should first deprecate them?
    - Yes. Deprecation will be a follow-up slated for v5.3.0.
2. Which button variants should we deprecate in v5.2.0 (if any)? (See [this az_quickstart search](https://github.com/search?q=repo%3Aaz-digital%2Faz_quickstart+btn-&type=code) for examples of button variants used in Quickstart.)
   - We will not deprecate any button variants in v5.2.0.
3. For the button variants in v5.2.0, what colors should be used for the `:focus-visible` outline?
   - For now, for button variants not included in the az-storybook component, the focus outline will match the button background color. We'll review the focus outlines with Marketing for a future release.
5. Is it acceptable to hold off on restricting button color variants based on the background color of their parent element?
   - At this time, these restrictions seem more appropriate to implement in Quickstart.
6. The solid blue button variant is not present in the Adobe XD designs. Is that variant still approved?
   - We'll check with Marketing and implement any changes in a future minor release.
7. Should we keep .btn-outline-info with a different text color from .btn-outline-sky?
   - Yes, we'll keep the azurite text color for .btn-outline-info.

## Related issues
 - #1956

## How to test

1. Visit https://review.digital.arizona.edu/arizona-bootstrap/issue/1956/docs/5.1/components/buttons/.
2. Confirm the class names for the new button styles are appropriate.
3. Confirm the appearance of all button colors and sizes (including `:hover` and `:focus-visible` styling) compared to the [button component in az-storybook](https://uaz-web.github.io/az-storybook/?path=/docs/components-button--buttons).

## Follow-up tasks
 - Review the button styles with Marketing and update according to the feedback received.
 - Update the Outline theme color variants section on the Buttons docs page to demonstrate that certain variants should be used on dark backgrounds.


djcelaya (changes requested):
## Review summary

Checked against #1956 and this PR's own description/questions.

**CI / build**
- ✅ Both CI checks pass (lint, build & deploy review site).
- ✅ Local containerized build (`scripts/build-container.sh`) and `docker run ... lint` both pass clean, matching CI.
- ✅ Diff is scoped to `scss/` and the docs page — no hand-committed `dist/**` or `package.json` version bump.

**Visual verification** (local dev server + headless screenshots)
- ✅ Padding, pill border-radius, Proxima Nova Condensed, font-weight 700, letter-spacing, and `.btn-lg` sizing all match the issue spec (accounting for the intentional whole-pixel-height adjustment explained in the PR description).
- ✅ `:focus-visible` outline (2px, offset, bloom by default / sky on `.btn-blue`) renders correctly and replaces the old box-shadow, confirmed with a rendered screenshot.
- ✅ New `.btn-sky`, `.btn-outline-sky`, `.btn-white-with-red`, `.btn-white-with-blue` variants render as expected; docs restructuring (Brand/Theme/Link variant sections, theme-colors loop) looks good.

**Two contrast regressions found (both introduced by this diff), computed against the actual compiled CSS:**

1. **`.btn-outline-info`** (`scss/override/_buttons.scss`): text/border color changed from `$azurite` to `$sky` (`#81ceeb`). On the white/light background it's shown against in the docs "Theme Color Variants" section, that's only a **~1.76:1** contrast ratio — well under WCAG AA's 4.5:1 (fails even the 3:1 large-text minimum). The old `$azurite` gave ~8:1. Note the equivalent brand-color outline variants (`.btn-outline-sky`, `.btn-outline-white`) are documented and demoed only on dark backgrounds — `.btn-outline-info` isn't, so as shipped it reads as safe for a plain background when it isn't.

2. **`.btn-success`** (via `scss/mixins/_buttons.scss`'s `az-button-variant`): the old `@if` chain special-cased `$background == $sky or $background == $leaf or $background == $light` for dark (`$midnight`) text. The rewrite splits `$sky` into its own `$blue`-text branch but drops `$leaf` entirely, so `.btn-success` (bg `$leaf` `#70b865`) now falls through to the default white text — confirmed in the compiled CSS (`.btn-success { color: #fff; }`, was previously dark). That's **~2.4:1** contrast, down from ~7.7:1. (Oddly, `.btn-success:hover` still gets `$midnight` via a separate untouched rule, so only the resting state regressed.) This looks like an unintentional drop during the `@if`/`@else if` refactor rather than a deliberate change.

Suggest restoring dark text for `$leaf`-backed buttons and reverting (or choosing a genuinely accessible color for) `.btn-outline-info`'s text color before merge — neither is caught by the lint/vnu-jar CI checks since they validate markup, not color contrast.

Everything else — build, lint, docs restructuring, new variants, focus states — looks solid. Nice work on the padding/line-height approach to get whole-pixel button heights.

danahertzberg:
For focus colors - not dire:

- Can the red button focus outline be red instead of bloom?
- Can the sky outline button focus color be sky instead of white?

scss/custom/_buttons.scss
```diff
@@ -19,12 +54,19 @@
   @extend .btn-outline-secondary;
 }
 
+.btn-outline-sky {
+  @extend .btn-outline-info;
+  --bs-btn-color: #{$sky};
+  --az-btn-focus-outline-color: #{$white};
```
danahertzberg, line 60:
```suggestion
  --az-btn-focus-outline-color: #{$sky};
```

scss/override/_buttons.scss
```diff
@@ -4,7 +4,35 @@
 // > BUTTONS
 //
 
-// >> Variants
+.btn {
+  --az-btn-focus-outline-color: var(--bs-bloom);
```
danahertzberg, line 8:
```suggestion
  --az-btn-focus-outline-color: var(--bs-red);
```

bberndt-uaz:
> For focus colors - not dire:
> 
> * Can the red button focus outline be red instead of bloom?
> * Can the sky outline button focus color be sky instead of white?

Those focus outline colors are defined in the Adobe XD designs and the storybook component.

danahertzberg:
> > For focus colors - not dire:
> > 
> > * Can the red button focus outline be red instead of bloom?
> > * Can the sky outline button focus color be sky instead of white?
> 
> Those focus outline colors are defined in the Adobe XD designs and the storybook component.

Thank you! No worries on this. We can talk with Tom at a future date about these focus colors.

danahertzberg (approved):

djcelaya (approved):

djcelaya:
Marketing approved this on 9/2.

djcelaya:
Side note: I am slightly less worried about the class names being perfect because we will need to change at least the new sky button in 5.3. For example, `.btn-sky` will become `.btn-rain` most likely. Instead of yanking them though, I think we could create class aliases to connect these and probably the many others that we'll need to change in the next release.

PR #2240, opened 2026-08-03
Closes #2086: Add support for serif style headings

djcelaya:
### Motivation

The arizona.edu redesign specifies an exact type scale (font-size, line-height,
font-weight, letter-spacing per breakpoint) for headings, and introduces Garamond Premier Pro as an opt-in serif style for headings. Proxima Nova remains the default font-family for all headings. See #2086.

### Changes

**1. New heading type scale (h1-h6, `.h1`-`.h6`)**
Applied the exact font-size/line-height/font-weight/letter-spacing spec for Proxima Nova headings:

| Level | Weight | Mobile (<992px) | Desktop (≥992px) |
| --- | --- | --- | --- |
| h1 | 600 | 48px / 56px | 64px / 74px |
| h2 | 600 | 38px / 46px | 52px / 62px |
| h3 | 600 | 30px / 38px | 40px / 50px |
| h4 | 600 | 28px / 36px (all widths) | — |
| h5 | 600 | 22px / 30px (all widths) | — |
| h6 | 600 | 18px / 26px (all widths) | — |

- h1-h3 are responsive at the `lg` (992px) breakpoint; h4-h6 are constant at all viewport widths (no mobile-specific spec was provided for h4-h6).
- Letter-spacing is explicitly `0` at all levels, matching the spec.
- Color remains AZ Blue by default; white-on-dark-background behavior is unchanged, handled by the existing `az-color-contrast()` mixin.
- Implemented via a new `$headings-scale` Sass map (`scss/custom/_variables.scss`) and an `@each` loop in `scss/override/_typography.scss`, rather than overriding Bootstrap's own `$h1-font-size`...`$h6-font-size` variables, to avoid unintentionally changing the `.fs-1`-`.fs-6` text utility classes (which are unrelated to actual headings but derive from the same Bootstrap variables) and to allow exact fixed sizing rather than Bootstrap's fluid RFS scaling.

**2. Opt-in serif (Garamond) container classes**
Two new container-level classes to opt any part of a page into serif headings:

- **`.az-headings-mixed`** — apply to `<body>` or any wrapper; switches only `h1`-`h3`/`.h1`-`.h3` to Garamond, leaving `h4`-`h6` on Proxima Nova.
- **`.az-headings-serif`** — apply to `<body>` or any wrapper; switches all of `h1`-`h6`/`.h1`-`.h6` to Garamond.

Both classes use the same font-size/line-height/font-weight/letter-spacing spec as Proxima Nova — only `font-family` changes.

[...]

PR #2292, opened 2026-09-01
Fixes #2291: .bd-title changes font-size on docs site

### LINKED ISSUE #2291: .bd-title changes font-size on docs site
There is a docs-specific class `bd-title` that sets a hard 3rem font size. We have noticed this applied at least to the H1 on the buttons docs page. This rule makes it so H1s appear smaller than H2s, which is very confusing to developers.

Image

I propose we remove the 3rem rule on the `.bd-title` class to allow the new headings sizes to come through.

djcelaya:
See #2291

_Note: `.bd-*` classes are Bootstrap docs specific styles._

## How to test
[Review site](https://review.digital.arizona.edu/arizona-bootstrap/bug/2291/)
1. Confirm that H1s with the `.bd-title` class are bigger than H2s. This regression was first observed on the [Buttons](https://review.digital.arizona.edu/arizona-bootstrap/bug/2291/docs/5.1/components/buttons/) docs page.
2. The `.bd-title` class no longer sets a font size, which should allow the new heading styles to come through. For H1s, the font size should be 64px and the line height should be 74px per #2240.

### DIFF
--- a/site/assets/scss/_content.scss
+++ b/site/assets/scss/_content.scss
@@ -103,7 +103,6 @@
 
 .bd-title {
   --bs-heading-color: var(--bs-emphasis-color);
-  @include font-size(3rem);
 }
 
 .bd-lead {
--- a/site/assets/scss/_content.scss
+++ b/site/assets/scss/_content.scss
@@ -103,6 +103,7 @@
 
 .bd-title {
   --bs-heading-color: var(--bs-emphasis-color);
+  @include font-size(3rem);
 }
 
 .bd-lead {
--- a/site/assets/scss/_custom-docs.scss
+++ b/site/assets/scss/_custom-docs.scss
@@ -285,6 +285,19 @@
 // > CONTENT
 //
 
+// >> Page title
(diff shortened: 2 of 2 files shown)


danahertzberg (approved):

trackleft (approved):

joeparsons (approved):

djcelaya:
Move this change to the custom-docs.scss file. Follow up with @bberndt-uaz if you run into trouble.

djcelaya:
Upstream pins `.bd-title` to a flat, RFS-scaled `3rem`. That beats Arizona
Bootstrap's own `h1` sizing, so docs page titles stayed at 3rem while every
other `h1` on the site grew to 4rem at `lg` — the mismatch reported in #2291.

Rather than just deleting upstream's declaration, this re-asserts the `h1`
values by reading them from `$headings-scale` (`scss/custom/_variables.scss`),
the same map that drives real `h1` elements in `scss/override/_typography.scss`.
If the type scale changes later, docs titles follow automatically instead of
drifting again.

bberndt-uaz (approved):

joeparsons (approved):

PR #2295, opened 2026-09-02
Suggested edits

trackleft:
## Summary

Generalizes the sticky-header scroll-offset fix that lived only in the docs site into a reusable feature of the Arizona Bootstrap library itself.

- **New `scss/custom/_scrolling.scss`** (library-level): defines `--az-scroll-behavior`, `--az-scroll-margin-top`, `--az-scroll-margin-bottom` custom properties (defaulting to `smooth` / `0px` / `0px`) and applies them as `scroll-behavior`/`scroll-margin-*` to focusable elements and headings (`a, button, input, select, textarea, h2, h3, h4, [tabindex="0"]`) inside `main`. Imported into `scss/arizona-bootstrap.scss` right after Bootstrap's functions import.
- **Accordion cleanup**: removed the accordion-specific `--az-accordion-anchor-scroll-offset` variable from `scss/custom/_accordion.scss`; `scss/override/_accordion.scss` now uses the new generic `--az-scroll-margin-top` / `--az-scroll-behavior` variables instead.
- **Docs site simplified**: `site/assets/scss/_scrolling.scss` no longer duplicates the selector list and rules (now inherited from the library) — it just overrides the three variables (`86px` top margin, `100px` bottom margin) to account for the docs site's sticky header.

**Net effect:** any consumer site with a sticky/fixed header can now fix focus-scroll-behind-header by overriding three well-named CSS variables, instead of the fix being baked into docs-only CSS and an accordion-only variable name that didn't reflect its actual general purpose.

### DIFF
--- a/scss/arizona-bootstrap.scss
+++ b/scss/arizona-bootstrap.scss
@@ -1,6 +1,8 @@
 // Upstream Bootstrap functions before anything else.
 @import "../node_modules/bootstrap/scss/functions";
 
+@import "custom/scrolling";
+
 // Arizona Bootstrap custom variables.
 @import "custom/variables";
 
--- a/scss/custom/_accordion.scss
+++ b/scss/custom/_accordion.scss
@@ -2,12 +2,6 @@
 // > Arizona Accordion custom component
 //
 
-// No offset by default - most consumer sites have no fixed/sticky header.
-// Sites that do have one should override this variable to match its height.
-:root {
-  --az-accordion-anchor-scroll-offset: 0;
-}
-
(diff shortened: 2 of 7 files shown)


zsaenz (approved):

djcelaya (approved):
This works great and seems like a good direction to go. We should probably update or remove this docs section as it is no longer accurate: https://review.digital.arizona.edu/arizona-bootstrap/issue-295-chris/docs/5.1/components/accordion/#scroll-offset

camikazegreen (approved):

bberndt-uaz (approved):

PR #2331, opened 2026-09-18
Fixes #2330: AZ Navbar Fullscreen: Mobile menu top rule overhangs the separators beneath it

### LINKED ISSUE #2330: AZ Navbar Fullscreen: Mobile menu top rule overhangs the separators beneath it
Below the `lg` breakpoint, the horizontal rule above the first mobile menu item is wider than the `.nav-item` separators below it, overhanging them on the right by the width of the scrollbar.

The rule is a `border-top` on `.navbar-az-fullscreen-nav-mobile-menu-container`, which spans that container's full width. The separators are `border-bottom` on `.nav-item`, which sit *inside* the scrolling column (`.navbar-az-fullscreen-modal-menu-nav-col`) and so give up width to the scrollbar. When the menu doesn't overflow there's no scrollbar, no width is lost, and the two line up — so this only shows once the menu scrolls.

### How to test

⚠️ **Resize a real browser window — do not use DevTools responsive/device mode.** Device emulation forces overlay scrollbars, which take no layout width and hide this entirely.

1. Open the [AZ Navbar Fullscreen example](https://review.digital.arizona.edu/arizona-bootstrap/main/docs/5.2/examples/navbar-az-fullscreen/).
2. Narrow the window to roughly 400–600px so the hamburger toggle appears, and open the site menu.
3. Make the window short enough that the menu list overflows and scrolls.
4. Compare the rule above the first item against the separators below it — the rule extends further right.

### Notes

Found while reviewing #2326. It is **not** caused by that PR — it reproduces on `main` at the same window size, and #2326 notes it as out of scope. Affects the secondary and tertiary submenu columns as well, since they use the same scrolling column class.

A fix is ready and will follow in a PR.

### On `main`

Image

### On `bug/2227`

Image

djcelaya:
Fixes #2330

Below the `lg` breakpoint, the rule above the first mobile menu item is a `border-top` on `.navbar-az-fullscreen-nav-mobile-menu-container`, so it spans that container's full width. The `.nav-item` separators below it live *inside* the scrolling column and give up width to the scrollbar, so the rule overhangs them on the right by the scrollbar's width.

Paint the rule inside the scrolling column instead, as a `::before` on `.navbar-az-fullscreen-modal-menu-nav-col`. Being in the scroller's content flow, it resolves to the same width the separators do — matching exactly, at whatever width the platform's scrollbar happens to be, and staying full width when the menu doesn't overflow. `position: sticky` keeps it pinned where the border used to sit rather than scrolling away with the content.

Two non-obvious details, which are why this isn't a one-liner:

- **The scroll shadow had to be nudged down.** `.navbar-az-fullscreen-nav-mobile-menu-container::before` is the fade gradient, and it's fully opaque at its top edge. It has no `top`, so it renders at its static position — which used to be *below* the container's `border-top`. Removing that border moved the gradient up by exactly 1px, directly onto the rule. It now carries a `margin-top` equal to the rule's height, restoring the clearance the border used to provide. Both read from `--az-navbar-fullscreen-mobile-menu-rule-height` so they can't drift apart.
- **The rule needs `z-index: 2`.** `.nav-toggle` is positioned at `z-index: 1`; at `z-index: 1` the rule lost the tie on document order, and the chevron drew over it — along with the bloom separator, which `.nav-toggle` carries up inside its stacking context.

Scoped to `.navbar-az-fullscreen-nav-mobile-menu-container`, which sits inside `#navbar-az-fullscreen-nav-mobile-col` (`col-12 d-lg-none`), so desktop never renders this subtree.

This is independent of #2326 and branches from `5.3.x` without it. The two touch `_navbar-fullscreen.scss` in different places; if #2326 lands first this will want a merge from the base before its diff reads cleanly.

## How to test

⚠️ **Resize a real browser window — do not use DevTools responsive/device mode.** Device emulation forces overlay scrollbars, which take no layout width and hide this entirely.

1. Open the [AZ Navbar Fullscreen example](https://review.digital.arizona.edu/arizona-bootstrap/bug/2330/docs/5.2/examples/navbar-az-fullscreen/) on the review site.
2. Narrow the window to roughly 400–600px so the hamburger toggle appears, and open the site menu.
3. Make the window short enough that the menu list overflows and scrolls.
4. The rule above the first item should now end exactly where the separators below it do.
5. Scroll the menu — the rule should stay pinned at the top rather than scrolling out of view.
6. Scroll until a chevron row passes under the rule — the line should stay unbroken across both the chevron ring and the red bloom separator.
7. Confirm the scroll shadow still fades in as you scroll.
8. Widen the menu until it no longer overflows — the rule should be full width, unchanged from `main`.
9. Drill into a submenu via a chevron and confirm the secondary/tertiary columns behave the same.
10. Widen back above 992px and confirm the desktop menu is unchanged.

Compare against [main](https://review.digital.arizona.edu/arizona-bootstrap/main/docs/5.2/examples/navbar-az-fullscreen/) at the same window size.

_Note: Since this is a CSS-only change, no additional follow up is needed in AZQS._

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bberndt-uaz (commented):
This solution does fix the issue with the top rule, but there seem to be some visual issues with the top of the scrollbar in Chrome and Firefox (but not Safari):

Image
<p></p>

The scroll shadow overlays part of the scrollbar. Also, with this PR, the shadow is shifted down 1 pixel, leaving a pixel line at the top of the scrollbar without the shadow.

However, there is a larger problem of extra right padding when a classic scrollbar is displayed at mobile breakpoints. Note the extra right padding in the second and third screenshots below.

#### `main` branch: Chrome DevTools with device type "Mobile" (how it should look on mobile devices)

Image

#### `main` branch: Chrome DevTools with device type "Desktop"

Image

#### `bug/2330` branch: Chrome DevTools with device type "Desktop"

Image
<p></p>

It appears that the attribute `style="padding-right: 15px;"` is being added [via JavaScript from upstream Bootstrap](https://github.com/twbs/bootstrap/blob/main/js/src/util/scrollbar.js#L40-L43) to AZ Navbar Fullscreen `.modal-*` elements when the modal is opened on desktop devices. I think if we're concerned about how the modal looks on desktop at mobile breakpoints, the overall right padding issue should be looked at as well.

According to [this article](https://www.smashingmagazine.com/2023/12/new-css-viewport-units-not-solve-classic-scrollbar-problem/#the-two-types-of-scrollbars), scrollbars are exclusively overlay on mobile platforms. In that case, I'd say these issues with showing the classic scrollbar in the mobile menu on desktop devices could be a low priority.

bberndt-uaz (approved):

mmunro-ltrr (approved):
I had difficulty with the before & after comparisons, but this might be a browser-specific issue. With Firefox 153.0 on Linux the problem wasn't showing up on the main branch example (though this PR also looked OK). But on a browser similar to Chrome (Chromium 153.0.8010.52) the initial breakage on main & fix on the PR branch did show up. The Arizona Digital BrowserStack account might still be active: should we run some more general checks with that?

djcelaya:
> This solution does fix the issue with the top rule, but there seem to be some visual issues with the top of the scrollbar in Chrome and Firefox (but not Safari):
> 
> Image
> The scroll shadow overlays part of the scrollbar. Also, with this PR, the shadow is shifted down 1 pixel, leaving a pixel line at the top of the scrollbar without the shadow.
> 
> However, there is a larger problem of extra right padding when a classic scrollbar is displayed at mobile breakpoints. Note the extra right padding in the second and third screenshots below.
> 
> #### `main` branch: Chrome DevTools with device type "Mobile" (how it should look on mobile devices)
> Image
> #### `main` branch: Chrome DevTools with device type "Desktop"
> Image
> #### `bug/2330` branch: Chrome DevTools with device type "Desktop"
> Image
> It appears that the attribute `style="padding-right: 15px;"` is being added [via JavaScript from upstream Bootstrap](https://github.com/twbs/bootstrap/blob/main/js/src/util/scrollbar.js#L40-L43) to AZ Navbar Fullscreen `.modal-*` elements when the modal is opened on desktop devices. I think if we're concerned about how the modal looks on desktop at mobile breakpoints, the overall right padding issue should be looked at as well.
> 
> According to [this article](https://www.smashingmagazine.com/2023/12/new-css-viewport-units-not-solve-classic-scrollbar-problem/#the-two-types-of-scrollbars), scrollbars are exclusively overlay on mobile platforms. In that case, I'd say these issues with showing the classic scrollbar in the mobile menu on desktop devices could be a low priority.

@bberndt-uaz, thank you for your review! I updated this branch with the color improvements from #2326 and the previously uncolored scrollbar is now looking much better. I am not seeing the "1px at the top of the scrollbar" problem anymore (though I'm sure I've seen it locally in working on these two).

Regarding the extra right spacing in the modal: to my knowledge, this space is reserved so that the modal does not shift if the non-modal page underneath also needs scrolling. Although it may look a little awkward, this extra space is expected (by me, at least) and appropriate given the current implementation. I won't go as far as to say that's the way it has to be, but I don't think it's worth investigating any more right now.

djcelaya:
> I had difficulty with the before & after comparisons, but this might be a browser-specific issue. With Firefox 153.0 on Linux the problem wasn't showing up on the main branch example (though this PR also looked OK). But on a browser similar to Chrome (Chromium 153.0.8010.52) the initial breakage on main & fix on the PR branch did show up. The Arizona Digital BrowserStack account might still be active: should we run some more general checks with that?

@mmunro-ltrr, thank you for your review! Sorry about the browser inconsistencies! I was toying around with BrowserStack and did confirm that Firefox 153 does not present the issue. I am still seeing this as a worthwhile improvement for browsers that do exhibit the issue.

PR #2332, opened 2026-09-18
Closes #1983: Use CSS variables wherever possible instead of hard-coded values (#2010)

### LINKED ISSUE #1983: Use CSS variables wherever possible instead of hard-coded values.

trackleft:
* Use CSS variables wherever possible instead of hard-coded values

- Convert all color references in scss/custom/ and scss/override/ to CSS custom properties
- Convert focus-ring Sass vars to CSS vars in _nav.scss, _carousel.scss, _dropdown.scss, _photo-gallery.scss
- Convert border-radius Sass vars to CSS vars in _nav.scss, _button-group.scss
- Convert $breadcrumb-item-padding-x to var(--bs-breadcrumb-item-padding-x)
- Move btn sizing/spacing/radius overrides from Sass vars to .btn CSS var block in _buttons.scss
- Convert btn box-shadow and focus-box-shadow to use CSS vars (--bs-white-rgb, --bs-black-rgb, --bs-focus-ring-width)
- Convert text-bg-* classes to use --bs-text-bg-color and --bs-text-bg-bg tokens with shared base rule
- Convert docs site SCSS (site/assets/scss/) brand color vars to CSS custom properties
- Add --bd-purple-light CSS var to docs :root block

The VR run on issue-1983 reported 29 failures. Four are false positives (docs code samples embed the branch name in asset URLs); the rest trace to CSS custom properties referenced from outside the scope that defines them. Unlike a Sass variable, an undefined var() resolves at runtime and silently invalidates the whole declaration, so neither stylelint nor the build caught these.

- Emit --bs-gray-*-rgb tokens. $table-striped-bg/$table-hover-bg reference rgba(var(--bs-gray-300-rgb), ...), but Bootstrap emits no -rgb triplet for the gray scale, so striped and hover table rows lost their background.

- Give --bs-nav-link-padding-y a 1.25rem fallback on .navbar-nav. The token it reads, --az-navbar-nav-link-padding-y, is only defined on .navbar-az, so every other navbar lost its vertical padding and shrank.

- Replace two more out-of-scope references with globally defined tokens of the same value: --bs-nav-tabs-link-active-bg (scoped to .nav-tabs) with --bs-secondary-bg for the active navbar item background, and --bs-nav-tabs-border-color with --bs-gray-200 for .nav-pills.flex-column dividers.

- Drop the --bs-breadcrumb-divider-color override, which replaced the brand #ef4056 with gray-600, and the unconsumed --bs-breadcrumb-divider-font-family token.

- Consume --bs-text-bg-color/--bs-text-bg-bg in .badge.badge-link. They were set but never read, so hover and focus styling was dead; !important is needed to beat the .text-bg-* utilities.

- Set --bs-nav-link-color, not --bs-nav-link-hover-color, for active navbar links in dark mode, and restore the azurite link color.

* Fix two more regressions found by a local VR round

Running the VR suite against this branch's review site cut the failures from 29 to 16, and showed two remaining differences.

- Active navbar links rendered white on their light active background, because `--bs-navbar-active-color` resolves to #fff on a dark navbar. Use $navbar-light-active-color (midnight, #03132e) as the baseline did.

- Table text rendered navy instead of black. `--bs-emphasis-color` was overridden to var(--bs-blue) in :root, and it feeds --bs-table-color among others. Its companion --bs-emphasis-color-rgb was left at 0, 0, 0, so the pair disagreed; dropping the override restores Bootstrap's #000.

* Fix active nav styling that CSS variables could not express

VR round 2 still showed active nav links rendering white on their light active background, unchanged from round 1. Bootstrap declares `color` directly on `.navbar-nav .nav-link.active`, which outranks the `--bs-nav-link-color` that `.nav-link` reads, so setting that custom property had no effect at all.

- Use `color` declarations for active navbar links, in both the default and dark colour modes. The values stay CSS variables, so the refactor's intent is kept; only the mechanism changes.

- Restore `background-color` and the border-radius on the generic `.nav-link.active` rule. Both were dropped in the migration, which left active nav pills and tabs with no background at all. `$nav-pills-border-radius` stays a Sass variable here because `--bs-nav-pills-border-radius` is only defined on `.nav-pills`, and this rule is generic.

Compiled output now matches the baseline exactly: #03132e on #e9ecef with var(--bs-border-radius).

* Restore the resting background on nav tabs

VR round 3 left one real regression: non-active tab links had lost their grey background. The migration substituted `--bs-nav-tabs-link-active-bg` for `$nav-tabs-link-bg`, but those are different tokens — the former is the active tab's background (white), the latter the resting background every tab shares (grey, #e9ecef). Use `$nav-tabs-link-bg`, which already resolves to var(--bs-secondary-bg).

* Keep utility nav links dark silver when active

Round 4 left one difference: active links in `.nav-utility` rendered midnight rather than dark silver.

This one was self-inflicted. Restoring `color` on the generic `.nav-link.active` rule (previous commit) gave it specificity 0,2,0, which outranks the `--bs-nav-link-color` that `.nav-utility` sets and `.nav-link` reads at 0,1,0. The baseline never hit this because it declared `color` directly on `.nav.nav-utility .nav-link` at 0,3,0.

Declare `color` on the active rule so it wins again. Non-active utility links are unaffected: nothing competes with the custom property there.

* Keep underline tabs on the page background

Restoring `background-color` on the generic `.nav-link.active` rule gave underline tabs a grey block they should not have: #2010 had also removed `background-color: $nav-tabs-link-active-bg` from
`.nav.nav-underline .nav-link.active`, which was invisible while the generic rule painted nothing, and became visible once it did.

Measured on the live review sites, `.nav-underline .nav-link.active` resolved to white on main, transparent on #2010, and #e9ecef here. Restoring the override brings it back to white.

An audit of every `.nav-link.active` on the navs and navs-tabs pages across all three branches found this to be the only such leak; every other active nav link already matched main.

---------

---------


trackleft:
Oops, this was a mistake pull request

